PR #4, opened 2016-11-11
Added output TFile to read_and_plot.py

bdorney:
read_and_plot.py will now make an output TFile having the distributions placed in an organized TDirectory structure.


jsturdy:
Replaced by #6

PR #6, opened 2016-11-25
Produce the output rootfiles and Mask all channels

gechen:


jsturdy:
Does this replace #4?

gechen:
yep

jsturdy:
OK! Thanks

PR #14, opened 2016-11-30
Update for new file format from pythonScript.py

gechen:


gechen:
@jsturdy

jsturdy (approved):
I think we can have some improvements, but since it's working we'll merge as is and iterate later

PR #20, opened 2016-12-01
Updated setTRIMDAC_MASK to take file as option

ReyerBand:


setTRIMDAC_MASK_FROM_FILE.py
```diff
@@ -22,11 +24,14 @@
 print testSuite.chipIDs
 
 for port in testSuite.presentVFAT2sSingle:
-    try:
-        trimDACfileList = open("TrimDACfiles.txt",'r')
-    except:
-        print "No TrimDACfiles.txt to specify paths to TRIM_DACS"
-        break
+    if len(trimfilelist) < 2:
```
jsturdy: Rather than this, set ```TrimDACfiles.txt``` as the default to this option, rather than the empty string ```""```

setTRIMDAC_MASK_FROM_FILE.py
```diff
@@ -22,11 +24,14 @@
...
+        try:
+            trimDACfileList = open("TrimDACfiles.txt",'r')
+        except:
+            print "No TrimDACfiles.txt to specify paths to TRIM_DACS"
+            break
+    else:
+        trimDACfileList = open(trimfilelist, 'r')
```
jsturdy: Should still do a ```try...except``` here, in case the file passed in is bad or doesn't exist

jsturdy (changes requested):
It appears that you modified your version which should have been removed.
After you update your repository (I opened a pull request to you) push your changes to your remote and this pull request will update

setTRIMDAC_and_Mask.py
```diff
@@ -23,9 +25,9 @@
 
 for port in testSuite.presentVFAT2sSingle:
     try:
```
jsturdy, line 27: I think the opening of this file can be done outside the VFAT loop.
That way you don't open and close this file 24 times

jsturdy (approved):
Looks good, comments can be addressed in future versions

PR #23, opened 2016-12-01
Took file read out of loop

ReyerBand:
Moved the trim file path read out of the VFAT loop as recommended, so we don't needlessly open a file 24 times.


jsturdy (approved):
Great!

PR #24, opened 2016-12-01
Update for each iEta region plotting

gechen:
@jsturdy

### DIFF
--- a/MaskChannels.py
+++ b/MaskChannels.py
@@ -6,15 +6,14 @@ Created on Wed Nov 23 19:02:34 2016
 @author: Geng
 
 """
-
-from ROOT import gROOT, gDirectory, TMultiGraph, TNamed, TLegend, TCanvas, TGraph, TH1F, TH2F, TFile, TDirectory #Classes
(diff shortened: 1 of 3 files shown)


MaskChannels.py
```diff
@@ -6,15 +6,14 @@
 @author: Geng
 
 """
-
-from ROOT import gROOT, gDirectory, TMultiGraph, TNamed, TLegend, TCanvas, TGraph, TH1F, TH2F, TFile, TDirectory #Classes
-from ROOT import kGreen, kYellow, kBlue, kRed #Colors
+from ROOT import *
```
jsturdy, line 9: Don't need both line 9 and 10.
Probably it's simplest to remove line 10, and change line 11 to be:
```gROOT.SetBatch(True)```

jsturdy (changes requested):
Can you make the change I suggest?
Then I'll approve and merge

MaskChannels.py
gechen: Updated.

jsturdy (approved):
Thanks!

PR #25, opened 2016-12-05
Update setTRIMDAC_and_Mask.py

jsturdy:
Fixed bug in only looping over first 127 channels


jsturdy:
Can you update this to also do a final threshold scan and set the threshold?

PR #26, opened 2016-12-05
Updating GEMDAQTestSuite

ReyerBand:


setTRIMDAC_and_Mask.py
```diff
@@ -11,9 +11,10 @@
                   help="GTX on the GLIB", metavar="gtx")
 parser.add_option("-f", "--file", type="string", dest="trimfilelist",
                   help="File containing paths to MASK_TrimDACs", metavar="trimfilelist", default="TrimDACfiles.txt")
-
-parser.add_option("-d", "--debug", action="store_true", dest="debug",
-                  help="print extra debugging information", metavar="debug")
+parser.add_option("-t", "--thresh", action="store_true", dest= "do_thresh",
+                  help="Do a threshold scan before/after setting trim", metavar="do_thresh")
+parser.add_option("-d", "--debug", action="store_true", dest= "debug",
+                  help="Debugging mode", metavar="debug")
```
jsturdy, line 18: Provide an option to use/not the ultra scan module, and then update the code accordingly

setTRIMDAC_and_Mask.py
```diff
@@ -77,3 +76,70 @@
...
+    startScanModule(testSuite.glib, options.gtx, useUltra = True, debug = options.debug)
+    UltraResults = getUltraScanResults(testSuite.glib, options.gtx, THRESH_MAX - THRESH_MIN + 1, options.debug)
+
+    if options.debug:
+        raw_input ("Press Enter to Continue")
+    print
+    print "Starting Preliminary Threshold Scan"
+    print
+    for n in testSuite.presentVFAT2sSingle:
+        # f = open("%s_Data_GLIB_IP_%s_VFAT2_%d_ID_0x%04x"%(str(Date),str(options.slot),n,testSuite.chipIDs[n]&0xffff),'w')
```
jsturdy: Is the writing out of these types of files really necessary for this script?
If so, the naming convention needs to be the same as in the updated code, and in fact, the whole block should be written as a function called here, rather than copied and pasted here

jsturdy (changes requested):
Pending confirmation that the ultra scan is working as desired, I'll approve this pull, but will request that the changes requested be implemented as soon as possible as well

jsturdy (approved):
Thanks

PR #27, opened 2016-12-11
Update to check empty files

gechen:


jsturdy (changes requested):
I would remove ```>>>>>>``` as they are present in git operations

gechen:
updated, @jsturdy

jsturdy (approved):
Thanks!

PR #42, opened 2017-03-14
cleanup and compatibility with updated cmsgemos

jsturdy:
* Modified ```import```s to use new ```cmsgemos``` paths
* Cleaned up some old code
* Intentionally crippled some obsolete scripts


jsturdy:
Merging with power

PR #43, opened 2017-03-16
Clean up and adding shifter scripts

cbravo135:


GEMDAQTestSuite.py
```diff
@@ -43,7 +43,6 @@ def __init__(self,slot,gtx,shelf=1,tests="",test_params=TEST_PARAMS(),debug=Fals
         """
         """
         self.slot   = slot
-        self.shelf  = shelf
```
jsturdy: I think something's missing here

takeScurves.py
```diff
@@ -0,0 +1,67 @@
...
+                  help="Number of I2C tests to perform on the VFAT2s (default is 100)", metavar="ni2c", default=100)
+parser.add_option("--ntrk", type="int", dest="ntrk",
+                  help="Number of tracking data packets to readout (default is 100)", metavar="ntrk", default=100)
+parser.add_option("--writeout", action="store_true", dest="writeout",
+                  help="Write the data to disk when testing the rate", metavar="writeout")
+parser.add_option("-d", "--debug", action="store_true", dest="debug",
+                  help="print extra debugging information", metavar="debug")
+
+(options, args) = parser.parse_args()
+
```
jsturdy: Should be cleaned up to use the qcoptions module

testing.py
```diff
@@ -20,3 +26,17 @@
...
+rpc_connect("eagle33")
+
+setTriggerSource(testSuite.glib,options.gtx,1)
+configureLocalT1(testSuite.glib, options.gtx, 1, 0, 40, 250, 0, options.debug)
+print 'Starting OH T1 controller'
+startLocalT1(testSuite.glib, options.gtx, options.debug)
+reg=getNode("GEM_AMC.OH.OH%i.T1Controller.TOGGLE"%options.gtx)
+print 'RPC register real_address ',hex(reg.real_address)
+print 'RPC register address ',hex(reg.address)
+#writeReg(reg,0x1)
```
jsturdy: This looks suspiciously like RPC related stuff, which should not be part of this PR

trimChamber.py
```diff
@@ -4,54 +4,64 @@
...
+from macros.fitScanData import fitScanData
 from gempython.tools.vfat_user_functions_uhal import *
-from fitScanData import fitScanData
 
 from qcoptions import parser
 
 parser.add_option("--path", type="string", dest="path", default="data",
                   help="Specify Output File Path", metavar="path")
+parser.add_option("-p", "--ptrim", type="float", dest="ptrim", default=4.0,
+                  help="Specify GEB (long/short)", metavar="GEBtype")
```
jsturdy: The help does not reflect this option parameter.
Also, suggest removing the single letter ```-p``` flag for future considerations

trimChamber.py
```diff
@@ -4,54 +4,64 @@
...
+  }
+
+if os.getenv('DATA_PATH') == None or os.getenv('DATA_PATH') == '':
+    print 'You must source the environment properly!'
+if os.getenv('BUILD_HOME') == None or os.getenv('BUILD_HOME') == '':
+    print 'You must source the environment properly!'
+
+
 from ROOT import TFile,TTree,TH1D,TCanvas,gROOT,gStyle,TF1
 import ROOT
```
jsturdy: Is this really needed?
If not, i.e., everything you need is handled in the previous line, don't just ```import ROOT```, otherwise, suggest 
```import ROOT as r```
to keep scoping, i.e., ```r.kRed``` rather than ```kRed``` for things that may have name clashes

**Review by jsturdy — CHANGES_REQUESTED** [review]
Needs a little more cleanup (in the comments)

File: testing.py [hunk 6]
```diff
@@ -1,18 +1,32 @@
-#!/bin/env python
-"""
-Script to take Scurve data using OH ultra scans
-By: Cameron Bravo (c.bravo@cern.ch)
-"""
+#!/usr/bin/env python
```
jsturdy: Missed this one earlier, should be ```#!/bin/env python```, I'll fix in merged version

procScurves.py
```diff
@@ -0,0 +1,50 @@
+import os
+import threading
+from optparse import OptionParser
+
+parser = OptionParser()
+
+parser.add_option("-r", "--run", type="string", dest="runname", default="2017.03.15-15.03.49.081532",
+                  help="Specify Run to Process", metavar="runname")
```
jsturdy, line 8: I would not specify a default here, and would instead do a check that the option was passed properly.

testing.py
```diff
@@ -2,11 +2,31 @@
...
+                  help="Number of I2C tests to perform on the VFAT2s (default is 100)", metavar="ni2c", default=100)
+parser.add_option("--ntrk", type="int", dest="ntrk",
+                  help="Number of tracking data packets to readout (default is 100)", metavar="ntrk", default=100)
+parser.add_option("--writeout", action="store_true", dest="writeout",
+                  help="Write the data to disk when testing the rate", metavar="writeout")
+parser.add_option("--tests", type="string", dest="tests",default="A,B,C,D,E,F",
+                  help="Tests to run, default is all", metavar="tests")
+
+parser.add_option("-d", "--debug", action="store_true", dest="debug",
+                  help="print extra debugging information", metavar="debug")
```
jsturdy: This also was reverted somehow...
Should be using the qcoptions import

testing.py
```diff
@@ -2,11 +2,31 @@
...
 
 (options, args) = parser.parse_args()
 
-test_params = TEST_PARAMS(namc=options.namc,
+test_params = TEST_PARAMS(nglib=options.nglib,
                           noh=options.noh,
                           ni2c=options.ni2c,
                           ntrk=options.ntrk,
```
jsturdy: This was reverted as well...

jsturdy (changes requested):
Seems a couple of more places got reverted, not sure what action did this

testing.py
jsturdy: Actually, as I requested more changes, why don't you make the change and include it in your update

cbravo135:
I checked this out from the central repo directly.

jsturdy (approved):

PR #44, opened 2017-03-21
Add masks to dead strips in trim script

cbravo135:


macros/fitScanData.py
```diff
@@ -8,23 +8,32 @@ def fitScanData(treeFile):
     inF = TFile(treeFile)
 
     scanHistos = {}
+    scanCount = {}
```
jsturdy, line 11: For code cleanup, you can use the ```nesteddict``` helper I added, like ```mkdir -p``` for ```dict```s

macros/anaUltraScurve.py
```diff
@@ -89,6 +89,8 @@
     myT.Branch( 'scurve_h', scurve_h)
     chi2 = array( 'f', [ 0 ] )
     myT.Branch( 'chi2', chi2, 'chi2/F')
+    Nev = array( 'f', [ 0 ] )
+    myT.Branch( 'Nev', Nev, 'Nev/F')
     pass
 
 vSum = {}
```
jsturdy, line 96: Can also use the ```nesteddict``` here (basically for all ```dict```s, though the benefit comes only when they're actually nested)

jsturdy (approved):
Nothing dealbreaker, but some improvements mentioned

PR #48, opened 2017-03-24
trimRange skip feature in trimChamber and confChamber tools

cbravo135:


anaAllChambers.py
```diff
@@ -2,15 +2,16 @@
 import threading
 from chamberInfo import chamber_config,GEBtype
 
-def launchScurveScan(link, filename,cType):
-  os.system("python $BUILD_HOME/vfatqc-python-scripts/macros/anaUltraScurve.py -i %s -f -t %s"%(filename,cType))
+def launchScurveScan(link, cName,cType):
+  os.system("python $BUILD_HOME/vfatqc-python-scripts/macros/anaUltraScurve.py -i data/%s/trimming/p_0.000000/current/SCurveData_trimdac0_range0.root -f -t %s"%(cName,cType))
```
jsturdy: Probably want the ```z``` parameter value to be configurable

trimAll.py
```diff
@@ -2,15 +2,29 @@
...
+  0:"GEMINI01L1",
+  1:"GEMINI01L2",
+  2:"GEMINI27L1",
+  3:"GEMINI27L2",
+  4:"GEMINI28L1",
+  5:"GEMINI28L2",
+  6:"GEMINI29L1",
+  7:"GEMINI29L2",
+  8:"GEMINI30L1",
+  9:"GEMINI30L2"
```
jsturdy: use the ```chamber_config```, unless there's a reason to redefine these here...

confChamber.py
```diff
@@ -9,7 +9,7 @@
 
 from qcoptions import parser
 
-parser.add_option("--filename", type="string", dest="filename", default="SCurveData_Trimmed.root",
+parser.add_option("--filename", type="string", dest="filename", default="",
```
jsturdy: Use ```default=None``` or don't set a default (which is effectively what ```default=None``` does)

confChamber.py
```diff
@@ -30,20 +30,19 @@
...
-filename = options.filename
-
 biasAllVFATs(ohboard,options.gtx,0x0,enable=False)
 print 'biased VFATs'
 writeAllVFATs(ohboard, options.gtx, "VThreshold1", options.vt1, 0)
 print 'Set VThreshold1 to %i'%options.vt1
 writeAllVFATs(ohboard, options.gtx, "ContReg0",    0x36,        0)
 
-#inF = TFile(filename)
+if options.filename != "":
```
jsturdy: Do a check on not being set to ```None``` and add a ```try```/```except``` to ensure the file exists.

trimChamber.py
```diff
@@ -26,6 +26,8 @@ def runCommand(cmd):
 
 from qcoptions import parser
 
+parser.add_option("--trimRange", type="string", dest="rangeFile", default="",
```
jsturdy: Same as above, don't set a default, then later do a check on being set, and a ```try```/```except``` to ensure the file exists.

trimChamber.py
```diff
@@ -101,69 +110,85 @@ def runCommand(cmd):
...
+        if(masks[vfat][ch]): continue
+        if(muFits_0[0][vfat][ch] - ptrim*muFits_0[1][vfat][ch] < sup[vfat] and muFits_0[0][vfat][ch] - ptrim*muFits_0[1][vfat][ch] > 0.1): 
+            sup[vfat] = muFits_0[0][vfat][ch] - ptrim*muFits_0[1][vfat][ch]
+            supCH[vfat] = ch
+    goodSup[vfat] = sup[vfat]
+    trimVcal[vfat] = sup[vfat]
+    trimCH[vfat] = supCH[vfat]
+    
+
+if rangeFile == "":
```
jsturdy: Same as above, do a check on being set, and a ```try```/```except``` to ensure the file exists.

jsturdy (changes requested):
As I like the simplicity of it, change all ```dict``` initializations from
```mydict = {}```
to
```mydict = ndict()```
importing the necessary library from ```gempython``` a la
```from gempython.utils.nesteddict import nesteddict as ndict```

jsturdy (approved):

PR #54, opened 2017-03-26
Reorganization of data and adding utime branch to ultraScurve TTree

cbravo135:


confAllChambers.py
```diff
@@ -18,7 +18,7 @@ def launchScurveScan(link,filename):
 
 for link in range(10):
     dataPath = os.getenv('DATA_PATH')
-    filename="%s/%s/trimming/p_0.000000/config/SCurveData_Trimmed.root"%(dataPath,chamber_config[link])
+    filename="%s/%s/trimming/z0.000000/config/SCurveData_Trimmed.root"%(dataPath,chamber_config[link])
```
jsturdy: This (z-value) should be configurable, no?

run_scans.py
```diff
@@ -7,49 +7,98 @@ def launchTestsArgs(tool, slot, link, chamber,vt1=None,vt2=0,perchannel=False,tr
   import datetime,os,sys
   import subprocess
   from subprocess import CalledProcessError
+  from chamberInfo import chamber_config
+
+  if os.getenv('DATA_PATH') == None or os.getenv('DATA_PATH') == '':
+    print 'You must source the environment properly!'
```
jsturdy: Add "DATA_PATH" not set to the message

run_scans.py
```diff
@@ -7,49 +7,98 @@ def launchTestsArgs(tool, slot, link, chamber,vt1=None,vt2=0,perchannel=False,tr
...
+    exit(0)
+  if os.getenv('BUILD_HOME') == None or os.getenv('BUILD_HOME') == '':
+    print 'You must source the environment properly!'
```
jsturdy: Add "BUILD_HOME" not set to the message

run_scans.py
jsturdy: I'd also move these checks to ```__main__``` where the check on the tool is done, otherwise you'll still submit all the jobs

run_scans.py
jsturdy: I'd also move these checks to ```__main__``` where the check on the tool is done, otherwise you'll still submit all the jobs

trimChamber.py
```diff
@@ -57,7 +59,8 @@ def runCommand(cmd):
 
 ohboard = getOHObject(options.slot,options.gtx,options.shelf,options.debug)
 
-dirPath = '%s/%s/trimming/z%f/%s'%(dataPath,chamber_config[options.gtx],ztrim,startTime)
+if options.dirPath == None: dirPath = '%s/%s/trimming/z%f/%s'%(dataPath,chamber_config[options.gtx],ztrim,startTime)
+else: dirPath = options.dirPath
 runCommand(["mkdir","-p",dirPath])
```
jsturdy: Let's migrate the ```runCommand``` wrapper out of ```trimChamber``` and use it as an imported function everywhere, i.e., ```run_scans```
I just pushed this to my PR on ```cmsgemos```, can be used a la:
```
from gempython.utils.wrappers import runCommand
...
runCommand(cmd)
```

jsturdy (changes requested):
Changes requested:
* Move environment checks to ```__main__```
* Add new usage of ```wrapper``` module for the commonly used ```subprocess.call``` blocks

trimChamber.py
cbravo135: I totally already put it into run_scans I just forgot to delete it here

confAllChambers.py
cbravo135: yeah, but we only have z=0.0 trim right now so eh

run_scans.py
```diff
@@ -9,13 +9,6 @@ def launchTestsArgs(tool, slot, link, chamber,vt1=None,vt2=0,perchannel=False,tr
   from subprocess import CalledProcessError
```
jsturdy, line 9: Instead of this import, let's 
```
from gempython.utils.wrappers import runCommand
```
and replace all the repeated blocks with
```
runCommand(cmd)
```

jsturdy (changes requested):
Just to move the ```subprocess.call``` blocks in ```run_scans.py``` to use ```runCommand```
Will make sure anywhere else using this syntax updates later

jsturdy (approved):
Good to go

PR #57, opened 2017-03-26
fix bug with timing of creation of log file

cbravo135:


run_scans.py
```diff
@@ -100,6 +99,7 @@ def launchTestsArgs(tool, slot, link, chamber,vt1=None,vt2=0,perchannel=False,tr
       pass
     print "executing", cmd
     sys.stdout.flush()
+    log = file("%s/scanLog.log"%(dirPath),"w")
```
jsturdy: ```log``` is referenced before this, could remove from the ```setupCmds``` block and just run before the ```preCmd``` block

jsturdy (approved):

PR #66, opened 2017-04-05
update to threshold scan tools

cbravo135:


buildConfig.py
```diff
@@ -0,0 +1,48 @@
...
+  exit(0)
+if os.getenv('CONFIG_PATH') == None or os.getenv('CONFIG_PATH') == '':
+  print 'You must source the environment properly, CONFIG_PATH is not set'
+  exit(0)
+if os.getenv('BUILD_HOME') == None or os.getenv('BUILD_HOME') == '':
+  print 'You must source the environment properly, BUILD_HOME is not set'
+  exit(0)
+
+
+def launchScurveScan(link,ztrim,cName,cType):
```
jsturdy: This function is being repeated a lot of places (of course with slight changes), would suggest to at the very least rename to something descriptive, an possibly abstract into a wrapper *if* possible

macros/PLOTTING/plot_scurves_by_thresh.py
```diff
@@ -20,21 +20,21 @@
 vfat = options.vfat
 strip = options.strip
 
-from ROOT import TLegend
+#from ROOT import TLegend
```
jsturdy: remove?

trimChamber.py
```diff
@@ -179,7 +179,7 @@
             pass
         pass
     except:
```
jsturdy: ```except Exception as e:```

jsturdy (changes requested):
Couple of minor things, and one related to exception handling (which may exist in other places, not visible in the diff)

jsturdy (approved):

PR #68, opened 2017-04-07
Fixed ultraThreshold, added options

jsturdy:
* Added some options
* Updated ultraThreshold
* updated run_scans
* some placeholders for future latency scans


run_scans.py
```diff
@@ -134,12 +157,15 @@ def launchTestsArgs(tool, slot, link, chamber,vt1=None,vt2=0,perchannel=False,tr
...
+                         [options.scanmax for x in range(len(chamber_config))],
+                         [options.nevts   for x in range(len(chamber_config))],
+                         [options.vt1     for x in range(len(chamber_config))],
+                         [options.vt2     for x in range(len(chamber_config))],
                          [options.perchannel for x in range(len(chamber_config))],
                          [options.trkdata for x in range(len(chamber_config))],
-                         [options.ztrim for x in range(len(chamber_config))],
-                         [options.config for x in range(len(chamber_config))]
+                         [options.ztrim   for x in range(len(chamber_config))],
+                         [options.config  for x in range(len(chamber_config))],
```
cbravo135, line 169: This last line of parameters shouldn't have a comma at the end

run_scans.py
```diff
@@ -63,13 +64,13 @@ def launchTestsArgs(tool, slot, link, chamber,vt1=None,vt2=0,perchannel=False,tr
...
     else:
       scanType = scanType + "/vfat"
-      pass
-    if trkdata:
-      cmd.append("--trkdata")
-      scanType = scanType + "/trk"
-      pass
-    else:
-      scanType = scanType + "/trig"
+      if trkdata:
```
cbravo135, line 67: We should also consider making a per channel trigger scan request impossible

qcoptions.py
```diff
@@ -21,4 +21,9 @@
                   help="[OPTIONAL] Run a shortened test after covers have been applied")
 parser.add_option("--ztrim", type="float", dest="ztrim", default=0.0,
                   help="Specify the p value of the trim", metavar="ztrim")
-
+parser.add_option("--scanmin", type="int", dest="scanmin",
```
cbravo135, line 24: I am not sure I am a fan of scanmin and scanmax being configurable, we should force the user to  scan over the whole range, and have a debug mode which does a shorter scan

cbravo135 (commented):
We should maybe discuss a bit what you added, not sure we want it to be easy to not take full scans since this would break the analysis a bit. If we update here, it should be updated to not break the analysis at the same time.

run_scans.py
jsturdy: Because python is awesome, this actually doesn't matter

qcoptions.py
jsturdy: Depending on the scan type, you don't necessarily want to scan the same range, and having these parameters removes the worse practice of hard-coding values

run_scans.py
jsturdy: Hmm, this was what I thought I had cleaned up, the logic currently goes:
```

    if perchannel:
      # do the per-channel scan setup
    else:
      # do the per VFAT scan setup, whether using tracking data or trigger data
```

run_scans.py
cbravo135: I figured you might care because it looks like all you really did here was clean up spacing.

ultraScurve.py
```diff
@@ -57,8 +61,8 @@
 
 ohboard = getOHObject(options.slot,options.gtx,options.shelf,options.debug)
 
-SCURVE_MIN = 0
-SCURVE_MAX = 254
+SCURVE_MIN = options.scanmin
```
cbravo135: Change these to be hardcoded to be the maximum range that works with the firmware so we don't have people not taking full scans.

ultraThreshold.py
```diff
@@ -71,8 +77,8 @@
 
 ohboard = getOHObject(options.slot,options.gtx,options.shelf,options.debug)
 
-THRESH_MIN = 0
-THRESH_MAX = 250
+THRESH_MIN = options.scanmin
```
cbravo135: These should also be hardcoded

cbravo135 (changes requested):
The scan range of ultraScurve.py and ultraThreshold.py should be hardcoded

cbravo135 (approved):

Issue #70, opened 2017-04-26
Make configurable the chambers passed into the system running scripts


jsturdy:
* ```run_scans.py``` and ```confAllChambers.py``` should be able to run over an arbitrary set of chambers and not expect a specific configuration
 * This can be implemented by a command-line switch *or* a config file, linked to an expected location

jsturdy:
* https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/master/confAllChambers.py#L42 should loop over a list of specified links
* https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/master/anaAllChambers.py#L34 should loop over a list of specified links
* https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/master/run_scans.py#L156 and following should be constructed using only a list of specified links

jsturdy:
Referenced by #71

cbravo135:
I guess this is done now? The way you did it didn't seem like the full feature you were looking for?

jsturdy:
Yeah, it's a stop-gap solution, requiring only one change, but still not necessarily the "best" solution

bdorney:
It seems that the "stop-gap" solution has gone on for ~3 months.  I suggest we close this issue since above line numbers no longer appear relevant.

PR #71, opened 2017-04-26
adding flexibility to select which chambers are used

jsturdy:
Just for review of what has been done, still needs to be tested

### DIFF
--- a/anaAllChambers.py
+++ b/anaAllChambers.py
@@ -16,7 +16,7 @@ def launchArgs(link,scandate,cName,cType,ztrim):
         return
 
 
-    cmd1 = ["python","$BUILD_HOME/vfatqc-python-scripts/macros/anaUltraScurve.py"]
+    cmd1 = ["python","%s/vfatqc-python-scripts/macros/anaUltraScurve.py"%(os.getenv("BUILD_HOME"))]
     cmd1.append("--infilename=%s"%(filename))
     cmd1.append("--fit")
     cmd1.append("--type=%s"%(cType))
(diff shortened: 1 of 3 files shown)


jsturdy:
references #70

anaAllChambers.py
```diff
@@ -1,41 +1,98 @@
...
+        print "No file to analyze. %s does not exist"%(filename)
+        return
+
+    cmd1 = ["python","%s/vfatqc-python-scripts/macros/anaUltraScurve.py"%(os.getenv("BUILD_HOME"))]
+    cmd1.append("--infilename=%s"%(filename))
+    cmd1.append("--fit")
+    cmd1.append("--type=%s"%(cType))
+
+    cmd2 = ["cp","%s/%s/scurve/%s/SCurveData/Summary.png"%(dataPath,cName,scandate),
+            "~/move/SCurveSummary_%s_ztrip%2.2f.png"%(cName,ztrim)]
```
cbravo135: typo here: ztrip->ztrim

anaAllChambers.py
```diff
@@ -1,41 +1,98 @@
...
+    import subprocess
+    import itertools
+    from multiprocessing import Pool, freeze_support
+    from chamberInfo import chamber_config,GEBtype
+    from gempython.utils.wrappers import envCheck
+    import glob
+
+    from qcoptions import parser
+
+    parser.add_option("--scandate", type="string", dest="scandate", default="current",
```
cbravo135, line 48: good idea, I was already thinking about this same thing

cbravo135 (commented):
So the plan is to comment out lines in the def'n of chamber config? Just one typo you should fix, then I will approve.

jsturdy:
one other question, in ```run_scans.py``` there was this line:
```    setupCmds.append( ["ln","-s",startTime,dirPath+"current"] )```
and it wasn't clear to me that it shouldn't be:
```    setupCmds.append( ["ln","-s",dirPath+startTime,dirPath+"current"] )```
clearly it's been working, so I know it works, so probably not a bug

cbravo135:
it is safer to use a relative path for the link than the full path, in the case you want to copy the whole directory somewhere.

cbravo135 (approved):

PR #75, opened 2017-05-02
New ultraScurve features and added NDF to fit tree

cbravo135:


macros/anaUltraScurve.py
```diff
@@ -115,6 +115,8 @@
     myT.Branch( 'scurve_h', scurve_h)
     chi2 = array( 'f', [ 0 ] )
     myT.Branch( 'chi2', chi2, 'chi2/F')
+    ndf = array( 'f', [ 0 ] )
```
jsturdy: Any reason this is defined as a ```float``` rather than some ```int``` type?

cbravo135: because I copied chi2, and I figured it didn't matter. TTrees play nicer with floats than ints. I can switch it if you want, I don't really care.

jsturdy: It's mostly about correctly describing things, as the performance hit of doing things "wrong" probably won't be noticed.
Things that are integer types should probably be stored as integer types.
In fact, looking now, I see there are other instances of using```float``` when ```int``` should be preferred (```Nev```, which changes types between input tree and output tree)

macros/fitScanData.py
```diff
@@ -54,7 +54,7 @@ def fitScanData(treeFile):
                 fitResult = scanHistos[vfat][ch].Fit('myERF','SQ')
                 #fitStatus = fitResult.Status()
                 fitChi2 = fitTF1.GetChisquare()
-                fitNDF = fitTF1.GetNDF()
+                fitNDF = int(fitTF1.GetNDF())
```
jsturdy: There's also no need to add this, as the return type is an integer type

jsturdy (changes requested):
Just two more changes related to this that I'll bother with here

macros/fitScanData.py
cbravo135: Isn't what I did safer anyway?

macros/anaUltraScurve.py
cbravo135: One could argue the "correct" way to do things is to have all the branches be floats and just cast all ints to floats. This is what I do in my analysis code. It is safer overall since integers are a subset of floats. Then you are also safe from funny type errors down the road since you know everything is a float down the road.

jsturdy (approved):

PR #77, opened 2017-05-08
Add channel range configuration options to ultraScurve.py

cbravo135:


ultraScurve.py
```diff
@@ -33,6 +37,14 @@
...
     exit(1)
     pass
+if options.chMin < 0 or options.chMin > 127:
+    print 'chMin must be in the range 0-127'
+    exit(1)
+    pass
+if options.chMax < 0 or options.chMax > 127:
+    print 'chMax must be in the range 0-127'
+    exit(1)
+    pass
```
jsturdy: * Should probably check that ```chMin < chMax```

jsturdy (changes requested):
Make sure that invalid ```chMin/chMax``` values aren't ever accepted

jsturdy (commented):
I'd prefer replacing the ```chMin/chMax``` checks with:
```
if not (0 <= chMin <= chMax < 128):
    print "chMin %d not in [0,%d] or chMax %d not in [%d,127]"%(chMin,chMax,chMax,chMin)
    exit(1)
    pass
```

jsturdy (approved):

Issue #18, opened 2016-12-01
Add VFAT Mask to Command Line Options Parser


bdorney:
Can we add a command line interface for the VFAT mask for the options parser in all python scripts in this package?

e.g. 

python2.7 pythonScript.py -s9 -g0 -vfatmask 7FFFFF

jsturdy:
What do you see as the purpose of ```vfatmask```?

bdorney:
Ideally in this script (or others) we could use vfatmask to choose which VFAT's the script would consider.

i.e. if someone wanted to run the S-Curve routine on VFATs 0-7, 9-23 they could use "vfatmask" to mask vfat 8.

jsturdy:
* Migrate common settings to a gempythonoptions package (name to be decided)
  * Provide import interface for this package 
  * Provide sanity checks on all parameter options
* New requested option ```--vfats``` should be a string e.g., ```./script.py --vfats 1,3,5-9,23``` which would be parsed and ```script.py``` would be run for the selected VFATs.
  * If the option is not specified, and the script expects it, the default should be to run over all VFATs

cbravo135:
Is this still relevant?

bdorney:
I could see it being beneficial during future QC8 activities, so I'd say "yes"

jsturdy:
Related to #70

bdorney:
Would also suggest that masked VFATs are skipped in python scan routines (e.g. ultra*.py scripts).  

For example when running on the setup in the cosmic stand (6 VFAT positions unreachable):

[GE11COSMICS@cosmicstandtif 20170616]$ /home/GE11COSMICS/Software/DAQ/vfatqc-python-scripts/ultraThreshold.py -s2 -g0 --shelf=3 --perchannel 2>&1 | tee runlog_thresh_trigData.txt
2017.06.16.15.00
Channel #0
GEM_AMC.OH.OH0.ScanController.ULTRA
FW scan mode       : 1
FW scan min        : 0
FW scan max        : 254
Ultra FW scan mask : 0x00f40400
FW scan channel    : 0
FW scan step size  : 1
FW scan n_triggers : 1000
FW scan status     : 0x00000000
No handlers could be found for logger "gempython.tools.vfat_user_functions_uhal"
An exception occurred gem.shelf03.amc02.optohybrid00: error on VFAT transaction (chip 10, ContReg3)

Adding:

            if (mask >> i) & 0x1: continue

[...]

PR #89, opened 2017-06-21
Updating Hot Chan Ident Algo and Adding VFAT Mask at cmd line 

bdorney:
Addressing:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/18

VFAT Mask now taken as a command line option for (implemented in qcoptions.py):
->ultraScurve.py
->ultraThreshold.py
->ultraLatency

Analysis macro: macros/anaUltraThreshold.py also takes vfat mask (required if ultraThreshold.py was run with a mask other than 0x0)

Additionally, hot channel identification algorithm in macros/anaUltraThreshold.py has been updated.  Algorithm now uses median absolute deviation with modified z-score (references as comments in qcutilities.py) to identify hot channels (outliers).

Optionally macros/anaUltraThreshold.py will update the channel mask and produced chConfig_MasksUpdated.txt to include hot channels identified by it and previous hot channels found in macros/anaUltraScurve.py if the user supplies a TFile with scurveFitTree TTree present.

Updates tested and confirmed to work:

http://cmsonline.cern.ch/cms-elog/991993

Requires: root_numpy-4.7.2 (some machines may still be on 4.5.X and would need to run 'pip install root_numpy --upgrade')

Now also addressing:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/91

New script anaAllChambersThresh.py will analyze all detectors stored under $DATA_PATH and transfer image files to $ELOG_PATH

Syntax:

anaAllChambersThresh.py --scandate=2017.06.15.15.23 --scandateTrim=2017.06.16.14.35 --ztrim=4

Here scandate is the date of the threshold scan and scandateTrim is the date of the trim run in which chConfig used for the threshold scan was determined.  Then ztrim is value of the trim range when trimChamber.py was called.  Note that scandate and scandateTrim default to "current."


.gitignore
```diff
@@ -9,3 +9,22 @@
...
+*.png
+*.pdf
+
+# Misc
+.DS_Store
+
+# ROOT Files
+*.root
+=======
+>>>>>>> 88c1a65bc71a9ed252b2bdfa802a7a27c233d6c8
```
jsturdy: Fix this failed merge

MaskChannels.py
```diff
@@ -190,12 +237,18 @@
...
+    if bool(VFAT2s_H) is False or bool(VFAT2s_S) is False:
+	continue
+=======
     VFAT2s_H.SetStats(kFALSE)
     VFAT2s_H.Draw('H SAME')
     Line_p0 = TLine(0,0,0,VFAT2s_H.GetMaximum())
     Line_p0.SetLineColor(4)
     Line_p0.Draw("SAME")
     legend.AddEntry(Line_p0, "cut if sigma = -1","L")
+>>>>>>> f69078e501fa5376408cd347169d20eaddad4cc2
```
jsturdy: fix this failed merge

MaskChannels.py
```diff
@@ -301,3 +357,4 @@
 
 
 
+>>>>>>> f69078e501fa5376408cd347169d20eaddad4cc2
```
jsturdy: fix this failed merge

macros/anaUltraScurve.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/env python
+#!/bin/env python2.7
 import os
```
jsturdy: Requires ```python2.7```?

macros/anaUltraThreshold.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/env python
+#!/bin/env python2.7
```
jsturdy: Requires ```python2.7```?

macros/anaUltraThreshold.py
```diff
@@ -7,7 +7,9 @@
 from PanChannelMaps import *
 from gempython.utils.nesteddict import nesteddict as ndict
 
-parser = OptionParser()
+from qcoptions import parser
+
+#parser = OptionParser()
```
jsturdy: remove

macros/anaUltraThreshold.py
```diff
@@ -27,23 +32,29 @@
...
+vfat_mask = options.vfatmask
+
 import ROOT as r
 r.gROOT.SetBatch(True)
 GEBtype = options.GEBtype
 inF = r.TFile(filename+'.root')
 outF = r.TFile(filename+'/'+outfilename, 'recreate')
 
-VT1_MAX = 250
+#VT1_MAX = 250
```
jsturdy: remove

macros/anaUltraThreshold.py
```diff
@@ -53,121 +64,215 @@
...
+            dict_vfatTrimMaskData[dataPt['vfatN']][dataPt[list_bNames[1]]]['trimDAC'] =  dataPt['trimDAC']
+	          pass
+    	  pass
+    except Exception as e:
+    	  print '%s does not seem to exist'%options.fileScurveFitTree
+    	  print e
+    	  pass
+    pass
+
+#Save Output
```
jsturdy: lots of strange indenting, try to configure editor to automatically convert tabs to spaces (looks like 4 is the default for these python files)

trimChamber.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/env python
+#!/bin/env python2.7
```
jsturdy: requires python2.7?

ultraLatency.py
```diff
@@ -1,8 +1,10 @@
-#!/bin/env python
+#!/bin/env python2.7
```
jsturdy: requires python2.7?

ultraScurve.py
```diff
@@ -1,7 +1,10 @@
-#!/bin/env python
+#!/bin/env python2.7
```
jsturdy: requires python2.7?

ultraLatency.py
```diff
@@ -139,6 +142,7 @@
     scanData = getUltraScanResults(ohboard, options.gtx, LATENCY_MAX - LATENCY_MIN + 1, options.debug)
 
     for i in range(0,24):
+        if (mask >> i) & 0x1: continue
```
jsturdy: this shouldn't be needed with lines 94 and 97 (should be confirmed)

ultraScurve.py
```diff
@@ -156,6 +162,7 @@
         startScanModule(ohboard, options.gtx, useUltra = True, debug = options.debug)
         scanData = getUltraScanResults(ohboard, options.gtx, SCURVE_MAX - SCURVE_MIN + 1, options.debug)
         for i in range(0,24):
+            if (mask >> i) & 0x1: continue
```
jsturdy, line 162: this shouldn't be needed with lines 128 and 158 (should be confirmed)

ultraThreshold.py
```diff
@@ -1,8 +1,11 @@
-#!/bin/env python
+#!/bin/env python2.7
```
jsturdy: requires python2.7?

ultraThreshold.py
```diff
@@ -137,6 +142,7 @@
         scanData = getUltraScanResults(ohboard, options.gtx, THRESH_MAX - THRESH_MIN + 1, options.debug)
         sys.stdout.flush()
         for i in range(0,24):
+            if (mask >> i) & 0x1: continue
```
jsturdy, line 145: this shouldn't be needed with lines 88 and 135 (should be confirmed)

ultraThreshold.py
```diff
@@ -105,6 +109,7 @@
             scanData = getUltraScanResults(ohboard, options.gtx, THRESH_MAX - THRESH_MIN + 1, options.debug)
             sys.stdout.flush()
             for i in range(0,24):
+            	if (mask >> i) & 0x1: continue
```
jsturdy, line 112: this shouldn't be needed with lines 88 and 102 (should be confirmed)

jsturdy (changes requested):
needs cleanup and some answers
possible check in cmsgemos for how the scan data is returned

macros/anaUltraThreshold.py
bdorney: Yes initialization of dict_vfatTrimMaskData at line 183 requires dict comprehension which was added in python2.7.  Does this implementation pose a problem?

macros/anaUltraThreshold.py
bdorney: Needed for option --vfatmask.  Can move specifically into this file however since it's already present in qcoptions makes sense to have it here.

If scan is taken with a vfatmask!=0 then trying to analyze that data with vfatmask == 0 causes seg fault.

Please advise.

macros/anaUltraThreshold.py
bdorney: In the analysis routine I don't see the problem with drawing out to VT1=255.

But yes was informed by cbravo that in the scan application there is a firmware bug that prevents going out to full range of this register.

macros/anaUltraThreshold.py
jsturdy: only if run on a machine without ```python2.7``` (post ```cc7``` switchover this won't be a problem), in which case, it should simply not run.
for ultimate portability, consider checking the python version, and initializing the dict depending on the version:
https://stackoverflow.com/questions/1747817/create-a-dictionary-with-list-comprehension-in-python#1747827

MaskChannels.py
bdorney: This is a Legacy routine.  From my understanding confChamber.py with --chConfig argument supersedes this script.

Suggest to remove MaskChannels.py completely (script was originally created at my request iirc).

MaskChannels.py
bdorney: see previous comment

MaskChannels.py
cbravo135: This file is obsolete anyway. It is better to mask channels using confChamber.py now.

macros/anaUltraThreshold.py
cbravo135: Where is this seg fault? We should be able to detect which vfats are present based on the data file. I think it would be better to make the analysis smart enough to see this and not need a --vfatmask option.

.gitignore
bdorney: Fixed, once resolved additional items will submit new pull request

macros/anaUltraThreshold.py
cbravo135: Again, we should be able to make the analysis robust to this. I agree with bdorney on this one.

macros/anaUltraThreshold.py
```diff
@@ -18,7 +20,10 @@
                   help="Make plots vs channels instead of strips", metavar="channels")
 parser.add_option("-p","--panasonic", action="store_true", dest="PanPin",
                   help="Make plots vs Panasonic pins instead of strips", metavar="PanPin")
-
+parser.add_option("--chConfigKnown", action="store_true", dest="chConfigKnown",
```
cbravo135, line 23: I think a better option would be to store this info in the data file produced by ultraThreshold.py. It is trivial to add these branches by just copying from ultraScurve.py.

macros/anaUltraThreshold.py
cbravo135: I agree, lets try to keep consistent indenting throughout the package.

macros/anaUltraScurve.py
bdorney: Running in python2.6.6 causes crash, maybe $PYTHONPATH is not set correctly in:

source $BUILD_HOME/cmsgemos/setup/paths.sh

However executing with python2.7 gives no problem.  What is the argument against requiring python2.7?  Installation recipe could just be updated include a local installation of python2.7 for all DAQ machines.

macros/anaUltraThreshold.py
```diff
@@ -53,121 +64,215 @@
...
 #Make a text file readable by TTree::ReadFile
-vt1 = []
-vt1new = []
-for i in range(0,24):
-    for j in range(VT1_MAX+1,0,-1):
-        #original method
-        if vSum[i].ProjectionY().GetBinContent(j+1) > 10.0:
-            print 'vt1 for VFAT %i found'%i
-            vt1.append(j+1)
+vt1 = {}
```
cbravo135: vt1new needs to be instantiated here. how did this get removed? how did this run without crashing when it appends to vt1new below.

macros/anaUltraThreshold.py
```diff
@@ -178,18 +283,42 @@
...
 
 print "trimRange:"
 print trimRange
 print "vt1:"
 print vt1
 
-txt.write("vfatN/I:vt1/I:vt1new/I:trimRange/I\n")
-for i in range(0,24):
-    txt.write('%i\t%i\t%i\t%i\n'%(i, vt1[i], vt1new[i], trimRange[i]))
+txt_vfat.write("vfatN/I:vt1/I:trimRange/I\n")
```
cbravo135, line 322: again, looks like vt1new got removed. It is a better algorithm, if anything we should remove vt1 and only use the vt1new algorithm. this part of the code seems to have been reverted somehow.

trimChamber.py
cbravo135: Again, this file is obsoleted by confChamber.py. Do we really care? I do think we should try to write code which is compatible with python and does not require python2.7 in general though.

ultraLatency.py
bdorney: Confirmed using --internal option on cosmic stand.

Removing

ultraScurve.py
bdorney: Commenting out line 165 generates exception:

[dorney@cosmicstandtif]~/scratch0/CMS_GEM/CMS_GEM_DAQ/testing/validation% ultraScurve.py -s2 -g0 --shelf=3 --vfatmask=0xF40400 2>&1 | tee scurveLog.txt
2017.06.21.17.45
Link 0 T1 controller status: 2
21-06-17 17:46:01.952104 [7fae214d8700] ERROR - Returned Header from URI "ipbustcp-2.0://192.168.1.42:60002", 0x2DF70104 ( transaction id = 0x00000DF7, transaction type = 0x00, word count = 1 ) had response field = 0x04 indicating an error (1 bytes into IPbus reply payload)
21-06-17 17:46:01.952148 [7fae214d8700] ERROR - Original sent header was 0x2DF7010F, for base address 0x00500E7A, 1 bytes into IPbus send payload
No handlers could be found for logger "gempython.tools.vfat_user_functions_uhal"
Channel #0
GEM_AMC.OH.OH0.ScanController.ULTRA
FW scan mode       : 3
FW scan min        : 0
FW scan max        : 254
Ultra FW scan mask : 0x00f40400
FW scan channel    : 0
FW scan step size  : 1
FW scan n_triggers : 1000
FW scan status     : 0x00000000
An exception occurred gem.shelf03.amc02.optohybrid00: error on VFAT transaction (chip 10, ContReg3)

Whereas running with line 165 uncommented produces expected behavior.  Note test stand used has dead I2C sectors for VFAT Pos 20-23 and no connector on GEB for Pos 10 and 18.

TL;DR: Line is needed.

ultraScurve.py
```diff
@@ -1,7 +1,10 @@
-#!/bin/env python
+#!/bin/env python2.7
 """
 Script to take Scurve data using OH ultra scans
 By: Cameron Bravo (c.bravo@cern.ch)
+
+Modified By: 
+	Brian Dorney (brian.l.dorney@cern.ch)
```
cbravo135: Looks like you modified how mask is set. If this is enough to add your name here, add sturdy too.

ultraScurve.py
cbravo135: This is needed because of lines 168, 169, and 170.

ultraThreshold.py
bdorney: Similarly to above comment regarding ultraScurve.py, if line is commented out generates exception.  If line is uncommented gives expected behavior.

Conclude line is needed.

macros/anaUltraScurve.py
jsturdy: Hmm, it's working correctly at p5 which has a default ```slc6``` (i.e., non-modern ```python```), definitely seems like an env error somewhere

macros/anaUltraThreshold.py
jsturdy: just meant for line 12, github wasn't smart enough this time...

macros/anaUltraThreshold.py
jsturdy: same here, only line 43

macros/anaUltraThreshold.py
bdorney: Tried already, implementation was not so trivial.  Containers used in ultraThreshold.py are indexed only by vfat position.  While a --perchannel option exists larger updates to the code would be needed for this solution.

Proposed solution I think is easier and requires less modification.  Additionally anaUltraThreshold.py works with or without this additional input due to these option flags.

macros/anaUltraThreshold.py
bdorney: It does not need to be instantiated here.

I changed vt1 from a list to a dict, by doing so you could store (key,value) items at the time of assignment.

Additionally vt1new is not present in the code (I forked from the central repository and don't have your changes).

macros/anaUltraThreshold.py
bdorney: The fork I have is from the central repo, I don't have your changes.

ultraScurve.py
bdorney: Lines removed, was just going through things spamming paste.

ultraThreshold.py
bdorney: Similarly to above comment regarding ultraScurve.py, if line is commented out generates exception. If line is uncommented gives expected behavior.

Conclude line is needed.

macros/anaUltraThreshold.py
jsturdy: * ```vt1new``` is obviated by the change that @bdorney is proposing
* be that as it may, one needs to ensure that one's code base is up to date before doing pull requests
  * standard operating procedure:
   * PR from central github to your fork
   * fetch from your fork to your working area
   * merge upstream changes into your feature branch

macros/anaUltraThreshold.py
bdorney: Yes agreed...however when I was using vim as an editor the indentation was consistent throughout.  When pushed to git it looks like the meaning of certain white space characters changed.  Will try to check this and resubmit.

trimChamber.py
bdorney: @cbravo135 I don't understand. This routine (trimChamber.py) is what is currently used in p5 to perform the trimming procedure is it not?

macros/anaUltraScurve.py
bdorney: could I have the output of:

echo $PYTHONPATH

From one of the machines at p5 where this works with the default python (2.6.X?)

trimChamber.py
cbravo135: yeah, nevermind, I thought this was as different file.

macros/anaUltraScurve.py
jsturdy:
```
/nfshome0/gempro/src/vfatqc-python-scripts:/nfshome0/gempro/src/cmsgemos:/lib:
```

macros/anaUltraThreshold.py
bdorney: Okay believe I have addressed this problem, will do a push later this afternoon

jsturdy:
One more note, if new package dependencies are being added, it's probably time to add a ```requirements.txt``` file, to list the package name and minimal/maximal compatible version
* https://pip.readthedocs.io/en/1.1/requirements.html
* ```pip``` style installation will *not* be available at p5, but workarounds can be found

bdorney:
Will add a requirements.txt file later this evening.  Made a new push after local testing, not sure how to trigger a review on the new material.

macros/anaUltraThreshold.py
```diff
@@ -53,143 +63,266 @@
...
+        #This dictionary has VFAT position as the key value, returns a structured numpy array
+        #dict_vfatTrimMaskData = {idx:initVFATArray(array_VFATSCurveData.dtype) for idx in np.unique(array_VFATSCurveData[list_bNames[0]])}
+        dict_vfatTrimMaskData = dict((idx,initVFATArray(array_VFATSCurveData.dtype)) for idx in np.unique(array_VFATSCurveData[list_bNames[0]]))
+        for dataPt in array_VFATSCurveData:
+            dict_vfatTrimMaskData[dataPt['vfatN']][dataPt[list_bNames[1]]]['mask'] =  dataPt['mask']
+            dict_vfatTrimMaskData[dataPt['vfatN']][dataPt[list_bNames[1]]]['trimDAC'] =  dataPt['trimDAC']
             pass
         pass
+    except Exception as e:
+        print '%s does not seem to exist'%options.fileScurveFitTree
```
jsturdy: very much doubt that any exception encountered here is due to a non-existent file...

jsturdy: very strongly doubt that an exception here is due to a non-existent file

pythonScript_ShorterDLYLessEvt.py
```diff
@@ -0,0 +1,732 @@
...
+            pass
+        f.close()
+        pass
+    pr.disable()
+    s = StringIO.StringIO()
+    sortby = 'cumulative'
+    ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
+    ps.print_stats()
+    print s.getvalue()
+
```
jsturdy: why did this file get added?

setTRIMDAC.py
```diff
@@ -506,10 +507,11 @@ def txtTitle(str):
...
-            regValue = (1 << 6) + int(trimDAC) #old version?
-            #print regValue
+            #print trimDAC
+            regValue = int(trimDAC) #correct version?
+            #regValue = (1 << 6) + int(trimDAC) #old version?
+	    #regValue = (1 << 6) - int(trimDAC)  #BLD version?
+            print regValue
             glib.set(regName, regValue)
 #We should make the S-curve optional, generally not a necessary check 
 #            glib.set('scan_reset', 1)
```
jsturdy: i really don't understand why all these files are getting modified

jsturdy (changes requested):
few small comments

setTRIMDAC.py
bdorney: I think this was old unmerged/uncommitted changes from when we were doing the slice test QC attempting to troubleshoot.

This script is superseded by confChamber.py with the --chConfig option.

Suggest to remove setTRIMDAC.py

pythonScript_ShorterDLYLessEvt.py
bdorney: I think this was some old unmerged/uncommitted changes from when we were doing the slice test QC and attempting to troubleshoot.

Most likely I copied the original "pythonScript.py" and did some ill-advised troubleshooting in December and the .gitignore was not set properly and this file was considered.

Will remove.

macros/anaUltraThreshold.py
bdorney: I don't follow, if options.fileScurveFitTree does not exist rp.root2array will generate an IOError since the option "warn_missing_tree" defaults to false:

http://scikit-hep.org/root_numpy/reference/generated/root_numpy.root2array.html#root_numpy.root2array

I don't see what other exceptions could be possible given that list_bNames is hard coded (based on the defined format of scurveFitTree).

bdorney: Looks like a duplicate comment, see response above

setTRIMDAC.py
jsturdy: Yet another cautionary tale about why one should *never* ```git commit -a```...

requirements.txt
```diff
@@ -0,0 +1,77 @@
...
+slip.gtk==0.2.20
+smbc==1.0
+SSSDConfig==1.13.3
+suds==0.4.1
+tornado==2.2.1
+urlgrabber==3.9.1
+urllib3==1.10.2
+virt-who==0.18
+virtinst==0.600.0
+yum-metadata-parser==1.1.2
```
jsturdy: there is no way all this is necessary, suggest ```root-numpy```, as you didn't add any other imports, right?

bdorney: Yeah for sure, that's why I was asking how I could go about seeing what was *actually* required on a blank machine.

But I think the critical packages beyond "slc6" default are given here:

http://cmsonline.cern.ch/cms-elog/992603
http://cmsonline.cern.ch/cms-elog/992570

Packages needed I believe are:

NumPy versions 1.6, 1.7, 1.8, 1.9, 1.10 (default on slc6 in python 2.6.6 is 1.4 and this is too old to work with root_numpy)
root-numpy-4.7.2

Will update to include just these.

ultraLatency.py
```diff
@@ -47,7 +49,7 @@
     uhal.setLogLevelTo( uhal.LogLevel.ERROR )
 
 from ROOT import TFile,TTree
-filename = options.filename
+filename = options.filename + "_shelf%i_s%i_g%i.root"%(options.shelf,options.slot,options.gtx)
```
jsturdy: This will not play nice with current infrastructure, where the structure is organized from top level as a given chamber and down.
Please revert for now, discussion will continue in #90

ultraScurve.py
```diff
@@ -48,7 +48,7 @@
     uhal.setLogLevelTo( uhal.LogLevel.ERROR )
 
 import ROOT as r
-filename = options.filename
+filename = options.filename + "_shelf%i_s%i_g%i.root"%(options.shelf,options.slot,options.gtx)
```
jsturdy: This will not play nice with current infrastructure, where the structure is organized from top level as a given chamber and down.
Please revert for now, discussion will continue in #90

ultraThreshold.py
```diff
@@ -34,7 +37,7 @@
     uhal.setLogLevelTo( uhal.LogLevel.ERROR )
 
 import ROOT as r
-filename = options.filename
+filename = options.filename + "_shelf%i_s%i_g%i.root"%(options.shelf,options.slot,options.gtx)
```
jsturdy: This will not play nice with current infrastructure, where the structure is organized from top level as a given chamber and down.
Please revert for now, discussion will continue in #90

jsturdy (changes requested):
Just needs an updated ```requirements.txt``` (i still see the busy version) and a reversion of the file naming convention

ultraLatency.py
bdorney: Reverting.

ultraScurve.py
bdorney: Reverting

ultraThreshold.py
bdorney: Reverting

bdorney:
Also expect an update to confChamber.py.  Will remove references to vt1new as this is no longer generated by macros/anaUltraThreshold.py

confChamber.py
```diff
@@ -81,8 +81,10 @@
         vfatTree.ReadFile(options.vfatConfig)
 
         for event in vfatTree :
-            print 'Set link %d VFAT%d VThreshold1 to %i'%(options.gtx,event.vfatN,event.vt1new+options.vt1bump)
-            writeVFAT(ohboard, options.gtx, int(event.vfatN), "VThreshold1", int(event.vt1new+options.vt1bump),0)
+            #print 'Set link %d VFAT%d VThreshold1 to %i'%(options.gtx,event.vfatN,event.vt1new+options.vt1bump)
+            print 'Set link %d VFAT%d VThreshold1 to %i'%(options.gtx,event.vfatN,event.vt1+options.vt1bump)
+            #writeVFAT(ohboard, options.gtx, int(event.vfatN), "VThreshold1", int(event.vt1new+options.vt1bump),0)
+            writeVFAT(ohboard, options.gtx, int(event.vfatN), "VThreshold1", int(event.vt1+options.vt1bump),0)
```
jsturdy: go ahead and remove the commented lines

requirements.txt
```diff
@@ -1,77 +0,0 @@
...
-slip.gtk==0.2.20
-smbc==1.0
-SSSDConfig==1.13.3
-suds==0.4.1
-tornado==2.2.1
-urlgrabber==3.9.1
-urllib3==1.10.2
-virt-who==0.18
-virtinst==0.600.0
-yum-metadata-parser==1.1.2
```
jsturdy: seems empty? did you remove completely? or is it just between pushes that i've checked

cbravo135: I looked at the file directly, it is empty.

ultraThreshold.py
```diff
@@ -37,7 +37,8 @@
     uhal.setLogLevelTo( uhal.LogLevel.ERROR )
 
 import ROOT as r
-filename = options.filename + "_shelf%i_s%i_g%i.root"%(options.shelf,options.slot,options.gtx)
+#filename = options.filename + "_shelf%i_s%i_g%i.root"%(options.shelf,options.slot,options.gtx)
```
cbravo135: remove this line

ultraScurve.py
```diff
@@ -48,7 +48,8 @@
     uhal.setLogLevelTo( uhal.LogLevel.ERROR )
 
 import ROOT as r
-filename = options.filename + "_shelf%i_s%i_g%i.root"%(options.shelf,options.slot,options.gtx)
+#filename = options.filename + "_shelf%i_s%i_g%i.root"%(options.shelf,options.slot,options.gtx)
```
cbravo135: remove this line

ultraLatency.py
```diff
@@ -49,7 +49,8 @@
     uhal.setLogLevelTo( uhal.LogLevel.ERROR )
 
 from ROOT import TFile,TTree
-filename = options.filename + "_shelf%i_s%i_g%i.root"%(options.shelf,options.slot,options.gtx)
+#filename = options.filename + "_shelf%i_s%i_g%i.root"%(options.shelf,options.slot,options.gtx)
```
cbravo135: remove this line

confChamber.py
bdorney: Will do

requirements.txt
bdorney: Yeah I see, regular expressions mistake, thanks for catching.  Fixing

ultraThreshold.py
bdorney: Done

ultraScurve.py
bdorney: Done

ultraLatency.py
bdorney: Done

macros/anaUltraThreshold.py
cbravo135: vt1new was in the central repo since it was being used at p5. If not this algorithm for vt1 should be changed to the much better algorithm that starts at the maximum bin and increases until I finds a the first zero bin. Which it looks like you use above in your hot channel masking code.

bdorney: Yeah some mistake on my part when forking, my bad...for sure will do better job of sync'ing in future.

In any event I think lines 134-139 are in the spirit of what we are looking for.

macros/anaUltraThreshold.py
cbravo135: We will also need to change the config files at p5 because right now the correct numbers are stored as vt1new. New configs should be built and placed in the correct spot when this finally gets deployed at p5.

bdorney: Could solve by adding a command line option to confChamber.py which allows the user to specify the string that correspond to the name of the "vt1" TBranch in the tree/text file

e.g.

--vt1BName=vt1
--vt1BName=vt1new

Makes script backwards compatible with default value given as vt1.  Please advise.

cbravo135: vt1new was always known to be just a temporary trial thing. Now that we are comfortable with it, might as well integrate it to just be vt1. The change you are suggesting is more complicated than you might think because it would have to be propagated properly to confAllChambers.py. We should try to keep things simple and not have lots of options users have to be aware of constantly. Lets try to build real solutions and less workarounds.

bdorney: Yup understand completely...was just suggesting a way to keep backwards compatibility.

Will leave it as simply "vt1"

requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy==1.10.4
+root-numpy==4.7.2
```
jsturdy: Last thing probably, but prefer a ```>=``` version check rather than a ```==``` as I doubt a fixed version is going to be mandatory

jsturdy (changes requested):

requirements.txt
bdorney: Done

bdorney:
Updated requirements.txt to use >= on version numbers

Added anaAllChambersThresh.py for analyzing threshold scan data from all chambers simultaneously.

anaAllChambersThresh.py
```diff
@@ -0,0 +1,142 @@
...
+        pass
+    if not foundDir:
+        print "Unable to find %s in output location specified: %s"%(options.scandate,searchPath)
+        exit(50)
+    else:
+        print "Found %s"%(options.scandate)
+        pass
+    
+    #Look for trim directory
+    searchPath  = "%s/GEMINI*/scurve/*"%(os.getenv('DATA_PATH'))
```
jsturdy: Are you looking for the trim directory here?  Or just some scurves?

macros/anaUltraThreshold.py
```diff
@@ -123,26 +125,37 @@
...
+from qcutilities import *
 import numpy as np
+import root_numpy as rp #note need root_numpy-4.7.2 (may need to run 'pip install root_numpy --upgrade')
+dict_hMaxVT1 = {}
+dict_hMaxVT1_NoOutlier = {}
 for vfat in range(0,24):
     if (vfat_mask >> vfat) & 0x1: continue
 
+    dict_hMaxVT1[vfat]          = r.TH1F('vfat%iChanMaxVT1'%vfat,"vfat%i"%vfat,255,0,255)
+    dict_hMaxVT1_NoOutlier[vfat]= r.TH1F('vfat%iChanMaxVT1_NoOutlier'%vfat,"vfat%i - No Outliers"%vfat,255,0,255)
```
jsturdy: Generally prefer centered bins, i.e., ```256,-0.5,255.5```

jsturdy (changes requested):
One question (which may require changes) and one change.

anaAllChambersThresh.py
bdorney: I am looking for the directory in which the TFile containing "scurveFitTree" is located.  This should correspond to the chConfig.txt file that was used in "confChamber.py" to setup the trimDAC's before running the threshold scan.

Explicitly if I wanted to analyze the threshold scan data taken:

http://cmsonline.cern.ch/cms-elog/991007

I would run with the scurveFitTree files obtained: http://cmsonline.cern.ch/cms-elog/990812

macros/anaUltraThreshold.py
bdorney: Will update.

anaAllChambersThresh.py
jsturdy: ok

jsturdy (approved):

jsturdy (changes requested):
remove hardcoded locations

bdorney:
What do you mean?  All locations appear to be based on a combination of environment variables:

$BUILD_HOME
$DATA_PATH
$ELOG_PATH

Detector names (e.g. GEMINImXLY) or scan dates.

anaAllChambersThresh.py
```diff
@@ -0,0 +1,149 @@
...
+    from chamberInfo import chamber_config
+    from gempython.utils.wrappers import runCommand
+
+    #print "launchArgs(%s,%s,%s,%s,%s)"%(link,scandate_scurve,scandate_thresh,cName,cType)
+
+    dataPath = os.getenv('DATA_PATH')
+    elogPath = "%s/%s"%(os.getenv('ELOG_PATH'),scandate_thresh)
+   
+    #filename_scurve = "%s/%s/scurve/%s/SCurveData/SCurveFitData.root"%(dataPath,cName,scandate_scurve)
+    filename_scurve = "%s/%s/trim/z4.000000/%s/SCurveData_Trimmed/SCurveFitData.root"%(dataPath,cName,scandate_scurve)
```
jsturdy: this

anaAllChambersThresh.py
```diff
@@ -0,0 +1,149 @@
...
+    if not foundDir:
+        print "Unable to find %s in output location specified: %s"%(options.scandate,searchPath)
+        exit(50)
+    else:
+        print "Found %s"%(options.scandate)
+        pass
+    
+    #Look for trim directory
+    #searchPath  = "%s/GEMINI*/scurve/*"%(os.getenv('DATA_PATH'))
+    searchPath  = "%s/GEMINI*/trim/z4.000000/*"%(os.getenv('DATA_PATH'))
```
jsturdy: this

anaAllChambersThresh.py
```diff
@@ -30,6 +31,11 @@ def launchArgs(link,scandate_scurve,scandate_thresh,cName,cType):
     cmd1.append("--fileScurveFitTree=%s"%(filename_scurve))
     cmd1.append("--vfatmask=0x0")
 
+    #print "================="
+    #for item in cmd1:
+    #    print item
+    #print "================="
+
```
jsturdy: this can go too

anaAllChambersThresh.py
jsturdy: i guess what i meant was remove the z value hard coding

anaAllChambersThresh.py
bdorney: If we make "z4.000000" a configurable parameter this should be sufficient yes?  The remainder of the structure seems to be required to fit the file structure setup on the gemdaq machine:

e.g.
filename_scurve = "%s/%s/trim/z4.000000/%s/SCurveData_Trimmed/SCurveFitData.root"%(dataPath,cName,scandate_scurve)

To:
+    filename_scurve = "%s/%s/trim/%s/%s/SCurveData_Trimmed/SCurveFitData.root"%(dataPath,cName,trimRange,scandate_scurve)

The "SCurveData_Trimmed/SCurveFitData.root" should remain fixed as it is in anaAllChambers.py line 15:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/master/anaAllChambers.py

anaAllChambersThresh.py
bdorney: Got it, will do

anaAllChambersThresh.py
bdorney: Done

bdorney:
Hard coded paths in anaAllChambersThresh.py have been removed, now the user should supply the "--ztrim" option to specify the correct trim filepath.

e.g.

anaAllChambersThresh.py --scandate=current --scandateTrim=current --ztrim=4

bdorney:
Finally based on feedback received updated hot channel definition.

Instead of starting at VT1_Max and working to first nonzero bin macros/anaUltraThreshold.py will start at the channel TH1::GetMaximumBin() and go to VT1_Max and identify the last non-zero bin.  This follows the old "vt1new" prescription as requested by @cbravo135.

anaAllChambersThresh.py
```diff
@@ -101,8 +93,8 @@ def launchArgs(link,scandate_scurve,scandate_thresh,cName,cType):
         pass
     
     #Look for trim directory
-    #searchPath  = "%s/GEMINI*/scurve/*"%(os.getenv('DATA_PATH'))
-    searchPath  = "%s/GEMINI*/trim/z4.000000/*"%(os.getenv('DATA_PATH'))
+    ztrimPath   = "z%s.000000"%(int(options.ztrim))
+    searchPath  = "%s/GEMINI*/trim/%s/*"%(os.getenv('DATA_PATH'),ztrimPath)
```
jsturdy: remove line 96 and change line 97 to:
```
searchPath  = "%s/GEMINI*/trim/%f/*"%(os.getenv('DATA_PATH'),options.ztrim)
```

anaAllChambersThresh.py
```diff
@@ -129,7 +121,8 @@ def launchArgs(link,scandate_scurve,scandate_thresh,cName,cType):
                                             [options.scandateTrim  for x in range(len(chamber_config))],
                                             [options.scandate  for x in range(len(chamber_config))],
                                             [chamber_config[x] for x in chamber_config.keys()],
-                                            [GEBtype[x]        for x in chamber_config.keys()]#,
+                                            [GEBtype[x]        for x in chamber_config.keys()],
+                                            [ztrimPath for x in range(len(chamber_config))]
```
jsturdy: here, just `options.ztrim`

anaAllChambersThresh.py
```diff
@@ -3,20 +3,17 @@
...
     from gempython.utils.wrappers import runCommand
 
-    #print "launchArgs(%s,%s,%s,%s,%s)"%(link,scandate_scurve,scandate_thresh,cName,cType)
-
     dataPath = os.getenv('DATA_PATH')
     elogPath = "%s/%s"%(os.getenv('ELOG_PATH'),scandate_thresh)
    
-    #filename_scurve = "%s/%s/scurve/%s/SCurveData/SCurveFitData.root"%(dataPath,cName,scandate_scurve)
-    filename_scurve = "%s/%s/trim/z4.000000/%s/SCurveData_Trimmed/SCurveFitData.root"%(dataPath,cName,scandate_scurve)
+    filename_scurve = "%s/%s/trim/%s/%s/SCurveData_Trimmed/SCurveFitData.root"%(dataPath,cName,zTrim,scandate_scurve)
```
jsturdy: Change line 16 to:
```
filename_scurve = "%s/%s/trim/%f/%s/SCurveData_Trimmed/SCurveFitData.root"%(dataPath,cName,zTrim,scandate_scurve)
```

jsturdy (changes requested):
Do you have comparisons between the results of your previous way of finding the max and the new way?

bdorney:
@jsturdy, no I do not have a comparison...the "new" way matches to the current P5 way, except in the rare case of a statistical fluctuation of the noise the two ways should yield similar results.

anaAllChambersThresh.py
bdorney: done

anaAllChambersThresh.py
bdorney: changed, thanks didn't realize %f was an option...now makes sense (float).

anaAllChambersThresh.py
bdorney: changed

jsturdy (approved):

jsturdy:
@mexanick, @cbravo135, I've done a merge to the `gemdev` account so if someone could test that our "usual" workflow still works, I think that we can merge this in fully
* Make sure that `confAllChambers` still works
* Make sure that the `run_scans` script still works (basically, just take a set of s-curves from this account)
Anything else we need to specifically test?

PR #92, opened 2017-06-28
AMC13 control

jsturdy:
* added control of some AMC13 settings into the `ultraLatency` script

### DIFF
--- a/chamberInfo.py
+++ b/chamberInfo.py
@@ -9,9 +9,7 @@ chamber_config = {
     6:"GEMINIm29L1",
     7:"GEMINIm29L2",
     8:"GEMINIm30L1",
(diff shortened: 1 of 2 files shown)


mexanick (commented):
Should we add here enabling L1As at CTP7 (just as a cross-check)?

jsturdy:
we could... though that's a simple thing to do on the CTP7
the AMC13 is the one that probably requires more intervention, and thus should be handled in the script

jsturdy:
OK, added it in here, though it needs to be properly tested, this should make things simple for the QC8 setup as there *should* be no intervention needed on the CTP7

mexanick (approved):
Looks correct to me

PR #96, opened 2017-07-11
Adding L1A counts printouts and other debug info

mexanick:
Adding L1A counts printouts and other debug info based on issue #127 of cms-gem-daq-project/cmsgemos:
https://github.com/cms-gem-daq-project/cmsgemos/issues/127


ultraLatency.py
```diff
@@ -188,10 +187,28 @@
...
+    amc13nL1A = (amc13board.read(amc13board.Board.T1, "STATUS.GENERAL.L1A_COUNT_HI") << 32) | (amc13board.read(amc13board.Board.T1, "STATUS.GENERAL.L1A_COUNT_LO"))
+    amcnL1A = getAMCL1Acount(amcboard)
+    ohnL1A = getLinkL1Acount(amcboard,options.gtx)
+    print "Initial L1A counts:"
+    print "AMC13: %s"%(amc13nL1A)
+    print "AMC: %s"%(amcnL1A)
+    print "OH%s: %s"%(link,ohnL1A)
     sys.stdout.flush()
-    scanData = getUltraScanResults(ohboard, options.gtx, LATENCY_MAX - LATENCY_MIN + 1, options.debug)
+    scanData = getUltraScanResults(ohboard, options.gtx, LATENCY_MAX - LATENCY_MIN + 1, options.debug, int(options.nevts), True)
```
cbravo135: Why do you hardcode the debug flag to true here?

ultraLatency.py
```diff
@@ -188,10 +187,28 @@
...
+    amc13nL1Af = (amc13board.read(amc13board.Board.T1, "STATUS.GENERAL.L1A_COUNT_HI") << 32) | (amc13board.read(amc13board.Board.T1, "STATUS.GENERAL.L1A_COUNT_LO"))
+    amcnL1Af = getAMCL1Acount(amcboard)
+    ohnL1Af = getLinkL1Acount(amcboard,options.gtx)
+    print "Final L1A counts:"
+    print "AMC13: %s, difference %s"%(amc13nL1Af,amc13nL1Af-amc13nL1A)
+    print "AMC: %s, difference %s"%(amcnL1Af,amcnL1Af-amcnL1A)
+    print "OH%s: %s, difference %s"%(link,ohnL1Af,ohnL1Af-ohnL1A)
+
+    for i in range(24):
+      print "Number of CRC errors for VFAT%s on link %s is %s"%(options.gtx, i, readRegister("GEM_AMC.OH.OH%d.COUNTERS.CRC.INCORRECT.VFAT%d"%(options.gtx,i)))
```
cbravo135: I do not see where you reset or read these registers before the scan starts. One of those two must happen.

cbravo135 (changes requested):
Reset or read CRC error counters before the scan starts.

ultraLatency.py
mexanick: it's not a debug flag, it is isLatency flag which is defaulted to be False, but here in ultraLatency script has to be set to True

ultraLatency.py
mexanick: These regs receive a reset in configureScanModule function. Please see corresponding changes in the cmsgemos repo

cbravo135: Did you check by hand that these registers work once?

mexanick: you mean this?
`eagle34 > read GEM_AMC.OH.OH0.COUNTERS.CRC.INCORRECT.VFAT0 

0x650280f0 r GEM_AMC.OH.OH0.COUNTERS.CRC.INCORRECT.VFAT0 0x003b2c95

eagle34 > write GEM_AMC.OH.OH0.COUNTERS.CRC.INCORRECT.VFAT0.Reset 1

0x00000001(1)   written to GEM_AMC.OH.OH0.COUNTERS.CRC.INCORRECT.VFAT0.Reset

eagle34 > read GEM_AMC.OH.OH0.COUNTERS.CRC.INCORRECT.VFAT0

0x650280f0 r    GEM_AMC.OH.OH0.COUNTERS.CRC.INCORRECT.VFAT0             0x00000000`

cbravo135 (approved):

ultraLatency.py
```diff
@@ -36,7 +36,7 @@
 parser.add_option("--fakeTTC", action="store_true", dest="fakeTTC",
                   help="Set up for using AMC13 local TTC generator", metavar="fakeTTC")
 
-parser.set_defaults(scanmin=153,scanmax=172,nevts=500)
+parser.set_defaults(scanmin=153,scanmax=172,nevts=500, amc13local=False, t3trig=False,fakeTTC=false,internal=False)
```
jsturdy: `store_true` options default to `False` (or `None` which evaluates to `False` if checked in an `if` statement), so this is unnecessary

ultraLatency.py
jsturdy: This should change based on the changes requested to the `cmsgemos` API

jsturdy (dismissed):

ultraLatency.py
```diff
@@ -188,10 +187,28 @@
     sys.stdout.flush()
     startScanModule(ohboard, options.gtx, useUltra=True, debug=options.debug)
     printScanConfiguration(ohboard, options.gtx, useUltra=True, debug=options.debug)
+    amc13nL1A = (amc13board.read(amc13board.Board.T1, "STATUS.GENERAL.L1A_COUNT_HI") << 32) | (amc13board.read(amc13board.Board.T1, "STATUS.GENERAL.L1A_COUNT_LO"))
+    amcnL1A = getAMCL1Acount(amcboard)
+    ohnL1A = getLinkL1Acount(ohboard,options.gtx)
+    print "Initial L1A counts:"
+    print "AMC13: %s"%(amc13nL1A)
+    print "AMC: %s"%(amcnL1A)
+    print "OH%s: %s"%(options.gtx,ohnL1A)
```
jsturdy: Looking at your elog, I would also slightly prefer the first print and read to be before the scan module is started.
* Granted, with high rate, before and after will have different values, but at the same time, *neither* will be "correct"

ultraLatency.py
```diff
@@ -94,12 +94,11 @@
 print(startTime)
 Date = startTime
 
-if options.amc13local:
-    import amc13
-    connection_file = "%s/connections.xml"%(os.getenv("GEM_ADDRESS_TABLE_PATH"))
-    amc13base  = "gem.shelf%02d.amc13"%(options.shelf)
-    amc13board = amc13.AMC13(connection_file,"%s.T1"%(amc13base),"%s.T2"%(amc13base))
-    pass
+import amc13
```
bdorney: Does this mean the "--amc13local" option will be depreciated?  If so suggest to remove from list of avaliable cmd options to limit option bloat.

ultraLatency.py
```diff
@@ -188,10 +187,28 @@
...
+    print "AMC13: %s, difference %s"%(amc13nL1Af,amc13nL1Af-amc13nL1A)
+    print "AMC: %s, difference %s"%(amcnL1Af,amcnL1Af-amcnL1A)
+    print "OH%s: %s, difference %s"%(options.gtx,ohnL1Af,ohnL1Af-ohnL1A)
+
+    for i in range(24):
+      print "Number of CRC errors for VFAT%s on link %s is %s"%(i, options.gtx, readRegister("GEM_AMC.OH.OH%d.COUNTERS.CRC.INCORRECT.VFAT%d"%(options.gtx,i)))
+
     sys.stdout.flush()
     for i in range(0,24):
         vfatN[0] = i
```
bdorney: What happens in these examples if VFAT X is missing or not responsive?

e.g. can we work in options.vfatmask somehow?

mexanick: nothing happens, that's not a VFAT register, printed value will be 0.

ultraLatency.py
mexanick: no, it doesn't. See
https://github.com/mexanick/vfatqc-python-scripts/blob/cbde5fe1e0bc6fe1324c5b1f10f61d3897ab6214/ultraLatency.py#L141

bdorney (changes requested):
Is options.amc13local depreciated by these changes?

For loops over vfats (particularly scanData[]) can we add a check on options.vfatmask?

ultraLatency.py
bdorney: Okay thanks for clarifying!

ultraLatency.py
bdorney: Got it, I guess your tests on QC8 stand didn't come up with any problems even though VFAT10 doesn't exist.

ultraLatency.py
cbravo135: I think we should also monitor the counter which counts the number of valid headers received by the OH for each vfat. This should equal the number of L1As, and if it doesn't, this would be evidence of unstable communication with VFATs.

cbravo135 (changes requested):
We should also use the valid data packet counters for each VFAT. We should make sure for each L1A, we get a valid data packet from all of the VFATs. The CRC check won't happen unless the OH sees a valid header come back from a VFAT.

jsturdy (commented):
OK, I think this is ready for testing

cbravo135 (approved):

Issue #97, opened 2017-07-12
anaUltraThresholds.py crashes when using the --vfatmask flag


lmoureaux:
Using anaUltraThresholds.py with the `--vfatmask=0xF40400` option leads to a crash at line 249:

    [...]
    Saving vt1Max distributions
    Traceback (most recent call last):
      File "/home/GE11COSMICS/Software/DAQ/vfatqc-python-scripts/macros/anaUltraThreshold.py", line 249, in 
        dict_hMaxVT1[vfat].Draw("hist")
    KeyError: 10

The crash occurs because the elements in dict_hMaxVT1 are created for non-masked VFATs only, but used for every VFAT. The '10' above is the index of the first masked VFAT.

See also http://cmsonline.cern.ch/cms-elog/996439

### Type of issue

- [x] Bug report
- [ ] Feature request

bdorney:
By changing the initialization of the dictionaries created in anaUltraThresholds.py you can remove the need for vfatmask entirely.

I've solved this on my develop branch...I'll have a PR "soon."

bdorney:
We can close this issue, it was addressed in:

https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/4

PR #98, opened 2017-07-12
anaUltraThresholds.py: Fix a KeyError when --vfatmask is used.

lmoureaux:
Fix for bug #97.

The elements in dict_hMaxVT1 were created for non-masked vfats only, but used
for every vfat. Use them only for non-masked vfats.


cbravo135 (approved):

macros/anaUltraThreshold.py
```diff
@@ -244,6 +244,7 @@
 r.gStyle.SetOptStat(0)
 print 'Saving vt1Max distributions'
 for vfat in range(0,24):
+    if (vfat_mask >> vfat) & 0x1: continue
     r.gStyle.SetOptStat(0)
     canv_vt1Max.cd(vfat+1)
     dict_hMaxVT1[vfat].Draw("hist")
```
jsturdy, line 250: Not knowing much of anything about this code, isn't it better to simply do something like (starting from line 246):
```python
for vfat in dict_hMaxVT1.keys():
    r.gStyle.SetOptStat(0)
    canv_vt1Max.cd(vfat+1)
    dict_hMaxVT1[vfat].Draw("hist")
...
```

If the code crashes when some data from some VFAT is missing, this is the better way to do the analysis, and should (probably) be changed in many places

jsturdy (changes requested):

macros/anaUltraThreshold.py
jsturdy: Should address issue #101

bdorney:
Let's close this PR, I've got an solution that discontinues the use of --vfatmask in analysis tools which I think is superior based on our discussion.  @cbravo135 is right, the analysis tools shouldn't need to know about the vfatmask.

macros/anaUltraThreshold.py
cbravo135: for vfat in dict_hMaxVT1.keys():
that only needs to be
for vfat in dict_hMaxVT1:
since python does these same thing in both cases. using a for in on a dict iterates over the keys.

Issue #99, opened 2017-07-12
Feature Request: Analysis Analog to run_scans.py


bdorney:
## Brief summary of issue
Suggest analogous script to run_scans.py but for analysis of scan data collected with python tools.  Suggest name be "ana_scans.py"

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

Providing command line arguments to run the analysis tools:
- macros/anaUltraScurve.py
- macros/anaUltraThreshold.py
- gem-plotting/latency/anaUltraLatency.py (suggest to rename from LatencyScanData.py to match naming convention)

Following in chamberInfo.py example also create an "anaInfo.py" file defining a dictionary of analysis type with analysis tool.

This should allow the DOC the ability to analyze any scan data taken and prepare configuration files and summary plots for posting to the elog.

Additionally suggest to separate analysis related options from qcoptions.py and create an anaoptions.py.  Also suggest to limit "option bloat" in these options scripts.  These scripts should only contain options that are common to multiple scripts (e.g. infilename or vfatmask).  Options that are specific to just one script should be migrated out of here and only move to those scripts.  This will improve the clarity of the help menus as many of the scripts right now dump a wall of text when calling help and many of the options provided take no action (e.g. --doLatency does nothing in ultraThreshold.py)

## Current Behavior

Right now we have:
- anaAllChambers.py
- anaAllChambersThresh.py

These are two separate tools that have to be maintained separately, in addition to the analysis scripts themselves.  This upgrade would depreciate these scripts and provide a single script for analysis that would serve as the central tool.

## Context (for feature requests)

DOC does not presently have the tools to seamlessly analyze data taken from all scans, proposed change would fix. 

This will also improve clarity of the help menu.

bdorney:
After discussion in the July 18th GEM Operations & Online SW Meeting we decided to migrate all analysis tools from vfatqc-python-scripts to gem-plotting-tools.

This issue will be taken care of there.

bdorney:
Addressed in:
https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/4

Closing

PR #100, opened 2017-07-12
Implementation of ana_scans.py

bdorney:
Addressing:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/99

Added "anaInfo.py" to define the analysis dictionary:
`
ana_config = {
        "latency":"anaUltraLatency.py",
        "scurve":"anaUltraScurve.py",
        "threshold":"anaUltraThreshold.py",
        "trim":"anaUltraScurve.py"
        }
`

This should be expanded whenever a new analysis tools/scan types are added and subsequent updates to launchAnaArgs() in ana_scans.py.

Added "anaoptions.py" to improve modularity of the code and decrease option bloat.  Here only cmd options used by scripts given in ana_config are shown.  Script specific analysis options are generally kept inside the corresponding analysis script.  This also caused some cmd options to migrate out of qcoptions.py.

Usage examples:

scurve analysis (tested, looks good)
`ana_scans.py --scandate1=current --ztrim=4 --anaType=scurve`

trim analysis (tested, looks good)
`ana_scans.py --scandate1=current  --ztrim=4 --anaType=trim`

latency analysis (gem-plotting/latency/anaUltraLatency.py for this to be successful)
`ana_scans.py --scandate1=current --anaType=latency`

threshold analysis (tested, looks good)
`ana_scans.py --scandate1=current --anaType=threshold`
`ana_scans.py --scandate1=current --scandate2=current --chConfigKnown --ztrim=4 --anaType=threshold`

Calling:
`ana_scans.py -h`

Should clarify things.  All scans require option "scandate1," analyzing threshold scans requires scandate2 (this takes the place of scandatetrim in anaAllChambersThresh.py and is more generically named).  In all cases scandate1 is the primary scandate.

This PR will depreciate:
- anaAllChambers.py
- anaAllChambersThresh.py

Functionality of these scripts is now present in ana_scans.py.  Suggest to remove "anaAllChambers*.py" to avoid clutter (will call "git rm" later).

This PR requires also:

https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/2

Changed the setup & installation recipe for this repo as well...now the user just needs to have:

`BUILD_HOME`
`DATA_PATH`
`ELOG_PATH`

Defined in the $SHELL and then call:

$BUILD_HOME/vfatqc-python-scripts/setup/paths.sh

This also resolves the one off issue of the detector mapping files not being created when attempting to run/analyze scan data.


bdorney:
Some thought is required on design of:

gem-plotting/latency/anaUltraLatency.py

Needs to be modified to match more closely with what macros/anaUltra*.py are doing (specifically make an output directory based on the input file name and place output files there).  During testing found that files made by anaUltraLatency.py are overwritten.  Will address this in the coming days but want to get this PR on the board.

anaAllChambers.py
```diff
@@ -46,11 +46,7 @@ def launchArgs(link,scandate,cName,cType,ztrim):
     from gempython.utils.wrappers import envCheck
     import glob
 
-    from qcoptions import parser
-
-    parser.add_option("--scandate", type="string", dest="scandate", default="current",
-                      help="Specify specific date to analyze", metavar="scandate")
-
+    from anaoptions import partser
```
jsturdy: typo?

ana_scans.py
```diff
@@ -0,0 +1,227 @@
...
+  from gempython.utils.wrappers import runCommand
+
+  dataPath  = os.getenv('DATA_PATH')
+  dirPath   = ""
+  elogPath  = "%s/%s"%(os.getenv('ELOG_PATH'),scandate1)
+    
+  print "Analysis Requested: %s"%(anaType)
+
+  #Build Commands
+  cmd = ["python"]
```
jsturdy, line 20: Make the script executable, add the location to the path, and don't mess with things this way (lines 20 and 24, 35, 60, 94).
Things will be cleaner and easier to re-organize

anaoptions.py
```diff
@@ -0,0 +1,24 @@
...
+parser.add_option("-i", "--infilename", type="string", dest="filename",
+                  help="Specify Input Filename", metavar="filename")
+parser.add_option("-p","--panasonic", action="store_true", dest="PanPin",
+                  help="Make plots vs Panasonic pins instead of strips", metavar="PanPin")
+parser.add_option("-o", "--outfilename", type="string", dest="outfilename",
+                  help="Specify Output Filename", metavar="outfilename")
+parser.add_option("--scandate1", type="string", dest="scandate1", default="current",
+                  help="Specify specific date to analyze", metavar="scandate1")
+parser.add_option("--scandate2", type="string", dest="scandate2", default=None,
+                  help="Specify supporting date to be used in analysis", metavar="scandate2")
```
jsturdy: Can something more descriptive than `scandate1` and `scandate2` be used?
I'd even prefer keeping `scandate` and just adding whatever the second one needs to be as a different option

qcoptions.py
```diff
@@ -1,23 +1,23 @@
...
+#parser.add_option("--writeout", action="store_true", dest="writeout",
+#                  help="Write the data to disk when testing the rate", metavar="writeout")
+#parser.add_option("--tests", type="string", dest="tests",default="A,B,E,F",
+#                  help="Tests to run, default is all", metavar="tests")
+#parser.add_option("--doLatency", action="store_true", dest="doLatency",
+#                  metavar="doLatency",
+#                  help="[OPTIONAL] Run latency scan to determine the latency value")
+#parser.add_option("--QC3test", action="store_true", dest="doQC3",
+#                  metavar="doQC3",
+#                  help="[OPTIONAL] Run a shortened test after covers have been applied")
```
jsturdy: Instructions were explicit: do not remove

anaAllChambers.py
bdorney: mmm, thought I fixed that...yes typo.

But plan to git rm this script

ana_scans.py
bdorney: Not sure I understand the comment.  The script is already executable.  For the path are you requesting a setup script be added?  If so shouldn't there just be a line added to:

`
cmsgemos/setup/paths.sh
`

What do you mean by "don't mess with things this way?"  I don't follow, could you clarify?

anaoptions.py
bdorney: Yes okay will do...was trying to keep it more general for future development but the point is well taken.

qcoptions.py
bdorney: These options seem to appear only in:
- pythonScript.py
- GEMDAQTestSuite.py

The first one is heavily depreciated and should probably be removed outright.

For the second one I see it could be useful in light of v3 commissioning.  Since these options are only really used in GEMDAQTestSuite.py why not move them there at say line 682?  I think this makes more sense from a design point of view.

e.g. why does "ultraThreshold.py" have the ni2c, QC3test, or doLatency options?  They don't do anything...seems to just be confusing for the user when calling the help menu.

ana_scans.py
jsturdy: You are setting it up in such a way that the execution of the script from the subshell is invoked via:
```bash
python <super long full path to script locatin> <options>
```
What I am requesting changes your operation procedure to do:
```bash
<script> <options>
```

>  If so shouldn't there just be a line added to:
`cmsgemos/setup/paths.sh`
> 

No, this repository is `vfatqc-python-scripts` and it is this location that needs to be added to the `$PATH`, *not* anything in `cmsgemos`

qcoptions.py
jsturdy: Point taken, however, then the solution should be to take the "scan" options out of the "qc" options parser and add a new `scanopts` parser object

ana_scans.py
bdorney: Okay I will try to change this.  

But I think there is still a misconception.  Yes I understand that `vfatqc-python-scripts` needs to be added to $PATH.  My question was _what is doing that_?

The only thing that I know that is the cms-gem-daq-project that sets up the environment is:

`
cmsgemos/setup/paths.sh
`

Are you asking:

`
$BUILD_HOME/vfatqc-python-scripts
`

Be added to the `$PATH` via `cmsgemos/setup/paths.sh`, which would require a separate PR in cmsgemos, or to include some new script that is included in `vfatqc-python-scripts` that the user should execute like paths.sh is executed?

qcoptions.py
bdorney: Okay a new parser object could be implemented.

But I think we clarify the design style.  For example `anaoptions.py` only includes those cmd line options that are common to multiple scripts while options that are only specific to script X are included in script X directly.  To me this approach seems easier to maintain than to have a Xoptions.py file that has an ever increasing set of parser objects specific to different scripts.

ana_scans.py
jsturdy: > Be added to the $PATH via cmsgemos/setup/paths.sh

*Definitely* not.  I have no interest in making `vfatqc-python-scripts` a dependency of `cmsgemos`
The currently it's probably done in a `.bashrc` or similar file (at P5 there's an env script for the `gempro` account)
> or to include some new script that is included in vfatqc-python-scripts that the user should execute like paths.sh is executed

Yeah, the best way is pretty simple: add a setup script to the repo to take care of setting up the environment for the package itself.

qcoptions.py
jsturdy: I agree, this set of options was originally functioning exactly like that.  At some point the scan options were added, and then some other things.  So now is not the worst time to get back to basics and make sure things are done coherently again

ana_scans.py
bdorney: Great okay, I'll work on a setup script for vfatqc-python-scripts...I have a prototype already.

qcoptions.py
bdorney: Kk I'll try to work on this implementation either tomorrow or next week.

ana_scans.py
bdorney: Okay the user just needs to ensure:

`BUILD_HOME`
`DATA_PATH`
`ELOG_PATH`

Are all defined in the `$SHELL` and then call:

`$BUILD_HOME/vfatqc-python-scripts/setup/paths.sh`

And then the environment will be setup correctly as we discussed.

bdorney:
I'd like to vote to either migrate all analysis options from this repo into gem-plotting repo, or move the latency analysis script into:

`vfatqc-python-scripts/macros`

Since it seems foolish to maintain a repo with one single file and then make it a dependancy of this script.

Also still need to work on updating anaUltraLatency.py in the gem-plotting repo

bdorney:
Closing this request based on our discussion in Tuesday's Online SW Meeting.

We will move all analysis tools into gem-plotting-tools

Issue #88, opened 2017-06-17
Moved from cms-gem-daq-project/cmsgemos#115


jsturdy:
Moved from cms-gem-daq-project/cmsgemos#115

bdorney:
Addressed by https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/102, closing.

Issue #86, opened 2017-06-07
Adding more configurable VFAT parameters 


jsturdy:
In order to use values other than the "defaults", there are several options:
* Have command line switches which modify the parameters object, a la:
```parameters.defaultValues["IComp"] = options.icomp```
* Have the ```vfatConfig_GEMINIm*.txt``` file contain any parameter that is different, and use a similar call as the previous option
```
vfatN/I:vt1/I:trimRange/I:IComp/I
0       20      1      90
1       18      0      90
2       20      1      90
3       23      1      90
4       39      1      90
5       23      1      90
6       52      3      90
7       52      2      90
8       22      2      90
9       20      1      90
10      19      1      90
11      27      1      90
12      16      1      90
13      17      1      90
14      35      1      90
15      58      3      90
16      18      1      90
17      20      1      90
18      17      1      90
19      14      1      90
20      20      1      90
21      15      1      90
22      50      2      90
23      75      3      90
```
* Add a per-chamber "defaults" in ```chamberInfo.py``` something like a new ```chamber_defaults```, or create more keys in the ```chamber_config``` dict, again passing them to the functions as above
```
chamber_defaults = {
    0: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
    1: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
    2: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
    3: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
    4: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
    5: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
    6: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
[...]

PR #102, opened 2017-07-15
Added --stepsize parameter

EklavyaFCB:
Addressing #88 

Updated `ultraLatency.py` file to enable the parameter --stepSize. 

This is was a straightforward implementation and required only:
- Adding another `parser.add_option()` with the necessarily information
- Adding the `options.stepSize` value in the configureScanModule function. The original function in cmsgemos/gempython/tools already had the required stepsize variable input.

Tested the updated script and can confirm with the following plots:
http://cmsonline.cern.ch/cms-elog/997121

Could/should add an if-statement in case the user inputs a stepsize larger than the difference of latencies (scanmax-scanmin).

Usage example (should add more parameters for a real scan):

`ultraLatency.py --scanmin=0 --scanmax=100 --stepSize=10`

Creates the `LatencyData_Trimmed.root` as normal.


EklavyaFCB:
@bdorney Here is the Pull Request of the stepsize feature implementation. Hope all is good.

bdorney:
Please add a check on stepSize:

> Could/should add an if-statement in case the user inputs a stepsize larger than the difference of latencies (scanmax-scanmin).

If:

`scanmin+stepSize > scanmax`

Then have the scan only execute the scan for scanmin and scanmax.  Explicitly change stepsize to be the difference between the two:

`stepSize = scanmax - scanmin`

EklavyaFCB:
Added

ultraLatency.py
```diff
@@ -54,6 +54,9 @@
     print("Invalid MSPL specified: %d, must be in range [1,8]"%(options.MSPL))
     exit(1)
 
+if (options.stepSize + options.scanmin > options.scanmax):
+    options.stepSize = options.scanmax - options.scanmin
+
```
bdorney: You should not try to change options.stepSize as this is cmd line entry from the user and may be required elsewhere.

Instead you should store options.stepSize as a local variable and then modify this variable accordingly

EklavyaFCB:
Modified.

Also added an else-statement in case the user doesn’t input a --stepwise parameter to not get an error in the configureScanModule function.

ultraLatency.py
```diff
@@ -35,6 +35,7 @@
                   help="Set up for using AMC13 T3 trigger input", metavar="t3trig")
 parser.add_option("--fakeTTC", action="store_true", dest="fakeTTC",
                   help="Set up for using AMC13 local TTC generator", metavar="fakeTTC")
+parser.add_option("--stepSize", type="int",dest="stepSize", help="Supply a step size to the latency scan from scanmin to scanmax", metavar="stepSize")
```
bdorney: Please make stepSize have a default value, set equal to 1.

ultraLatency.py
```diff
@@ -53,6 +54,11 @@
     print("Invalid MSPL specified: %d, must be in range [1,8]"%(options.MSPL))
     exit(1)
 
+if options.stepSize:
+    step = options.stepSize
+    if (step + options.scanmin > options.scanmax):
+        step = options.scanmax - options.scanmin
+
 if options.debug:
     uhal.setLogLevelTo(uhal.LogLevel.INFO)
```
bdorney: This case structure at line 57 is not needed.  Remove, only the case structure at line 59-60 should be used.   

Additionally you need to assign the value of step before the case structure, otherwise you will have a scope issue, e.g. at line 193 `step=0` because of this scope issue.

ultraLatency.py
```diff
@@ -53,6 +54,11 @@
     print("Invalid MSPL specified: %d, must be in range [1,8]"%(options.MSPL))
     exit(1)
 
+if options.stepSize:
```
cbravo135: There is no reason for this check if you have a default value. What is it you are trying to check here anyway?

cbravo135: I would check step > 0 and that seems like the only thing you need. We also should propogate this change into run_scans.py before we can merge this change.

cbravo135 (changes requested):
Please propagate this change to include this option in run_scans.py

ultraLatency.py
```diff
@@ -53,6 +53,14 @@
     print("Invalid MSPL specified: %d, must be in range [1,8]"%(options.MSPL))
     exit(1)
 
+if options.stepSize <= 0:
+    print("Invalid stepSize specified: %d, must be in range [1, %d]"%(options.stepSize, options.scanmax-options.scanmin))
+    exit(1)
+
+step = options.stepSize
```
cbravo135, line 62: why have this line?

ultraLatency.py
```diff
@@ -182,6 +190,7 @@
         pass
     configureScanModule(ohboard, options.gtx, mode, mask,
                         scanmin=LATENCY_MIN, scanmax=LATENCY_MAX,
+                        stepsize=step,
```
cbravo135: Just use options.stepSize directly here, you only use it once...

cbravo135 (dismissed):
Please propagate the changes to run_scans.

ultraLatency.py
EklavyaFCB: Yes, but we want the value of stepsize to be at max equal to scanmax-scanmin. So we assign to a variable the value of options.stepsize, rather than modify options.stepsize directly as the latter is a command line entry and could be required elsewhere as Brian pointed out. 

Therefore in the function, the variable with the updated value should be used, not whats input in the command line.

cbravo135: Sure, this is not nearly as important as this change being propagated to run_scans. We can not approve this PR until this happens.

EklavyaFCB: Yeah, sure, I didn't mean for all the commits to send PR notifications every time. Just needed to update my repo to download it in a different location. 

I implemented some changes run_scans.py but couldn't test them yet (my $DATA_PATH didn't have permissions yet). Soon ...

bdorney: `$DATA_PATH` is a `$SHELL` variable that you set.  You can change the directory it is set too, and based on my earlier comments you should.

EklavyaFCB (commented):
There were some merge conflicts with my version of the files vs the ones Brian updated in the last few days. Have solved the conflicts so the merge shouldn't be an issue.

ultraLatency.py
EklavyaFCB: For lines 56-58: in case the user inputs a negative stepSize, we should print an error line, like with scanmin/max, MSPL and vt2 above.

For line 60: as mentioned earlier, to use a variable of hold the value of stepSize rather than using the direct CLI value, in order to calculate the difference between the scanmax and min if necessary.

run_scans.py
```diff
@@ -106,6 +107,12 @@ def launchTestsArgs(tool, shelf, slot, link, chamber, scanmin, scanmax, nevts,
     cmd.append( "--scanmax=%d"%(scanmax) )
     cmd.append( "--nevts=%d"%(nevts) )
     cmd.append( "--throttle=%i"%(throttle) )
+    if stepSize > 0:
+      step = stepSize
+      if (step + scanmin > scanmax):
+        step = scanmax - scanmin
+      cmd.append( "--stepSize=%d"%(step) )
+      pass
```
bdorney: These lines in run_scans.py are unnecessary. You already coded this check in ultraLatency.py.

You should only be passing the parameter here

run_scans.py
```diff
@@ -3,7 +3,8 @@
 def launchTests(args):
   return launchTestsArgs(*args)
 
-def launchTestsArgs(tool, shelf, slot, link, chamber, scanmin, scanmax, nevts,
+
+def launchTestsArgs(tool, shelf, slot, link, chamber, scanmin, scanmax, stepSize, nevts,
```
bdorney: What is the default value that stepSize will take?

Ensure that it is 1.

bdorney (changes requested):
Changes requested. Cross check what default value is being passed to launchArgs for stepSize.

Additionally launchArgs should only pass parameters to the tool of interest; so please revise lines 110-115 of run_scans.py

run_scans.py
```diff
@@ -107,10 +107,7 @@ def launchTestsArgs(tool, shelf, slot, link, chamber, scanmin, scanmax, stepSize
     cmd.append( "--scanmax=%d"%(scanmax) )
     cmd.append( "--nevts=%d"%(nevts) )
     cmd.append( "--throttle=%i"%(throttle) )
-    if stepSize > 0:
-      step = stepSize
-      if (step + scanmin > scanmax):
-        step = scanmax - scanmin
+    if stepSize:
       cmd.append( "--stepSize=%d"%(step) )
```
bdorney: `stepSize` is not boolean but integer, example, how is `stepSize = 20` treated?  Please change to something like `stepSize > 0`

Also `step` looks like it has the value of 0 here, why use another variable at all? You have `stepSize`, use that.

qcoptions.py
```diff
@@ -8,7 +8,9 @@
                   help="Minimum value of scan parameter", metavar="scanmin", default=0)
 parser.add_option("--scanmax", type="int", dest="scanmax",
                   help="Maximum value of scan parameter", metavar="scanmax", default=254)
-parser.add_option("--vfatmask", type="int", dest="vfatmask", 
-                  help="VFATs to be masked in scan & analysis applications (e.g. 0xFFFFF masks all VFATs)", metavar="vfatmask", default=0x0)
+parser.add_option("--vfatmask", type="int", dest="vfatmask",
+		              help="VFATs to be masked in scan & analysis applications (e.g. 0xFFFFF masks all VFATs)", metavar="vfatmask", default=0x0)
+parser.add_option("--stepSize", type="int", dest="stepSize",
```
bdorney: Right now are run_scans.py and ultraLatency.py the only ones that use the `stepSize` option?  I think yes, please keep `stepSize` option hard coded in only these two scripts to avoid confusion.  For example if you call:

`ultraThreshold.py -h`

You'll see the `stepSize` option right now but it doesn't do anything in `ultraThreshold.py.`

Please update.

Also, keep the options menu's alphabetized for readability.

bdorney (changes requested):

run_scans.py
EklavyaFCB: Ack, I did have options.stepSize there, but b/c of the git push CLI issues, I updated to code on the web GUI, and missed that one.

bdorney:
Please organize time for testing on the cosmic stand then upload the elog short link of the results here

EklavyaFCB:
Tests for run_scans.py and ultraLatency.py can be seen here:

http://cmsonline.cern.ch/cms-elog/1000339

The only small remaining issue is the `[None]` that gets printed before the `Normal termination`. It creates the .root files in `GE11-VI-L-CERN-0001/latency/trk/current` fine though, then you just have to run the analysis on them, using anaUltraLatency.py -i in this case with the correct path of the .root file.

EklavyaFCB:
The last commit was for a different pull request (#86) that I'd prepared. Just saw that you can't make another PR until a previous one is fully approved and merged. In which case I can remove the changes in the latest commit if required. Please let me know.

jsturdy:
See [this](http://nvie.com/posts/a-successful-git-branching-model/) for a description of the branching model that we try to use (though not as rigorously in this repo yet... (not to name names)
Basic outline:
* Start from your branch which is up-to-date with the upstream branch you're developing a feature for
* Create a new branch `my-great-new-feature`
* Add commits to *this* feature branch related to the feature you are developing
* Push this branch to your remote
* Create a PR from `my-great-new-feature` to the upstream branch that your feature is intended for

If you happen to be developing two independent features, you do the exact same thing for the second feature.  In this way, you won't have the issue that you're encountering now, and all PRs will be more cleanly managed.
There may be problems if your two features depend on one another, but this can easily be accounted for by making your dependent feature a new branch off of your feature development branch

EklavyaFCB:
Thanks Jared, appreciate the advice. Have removed the updates in the latest commit, so this PR should be good to go based on the e-log I posted earlier.

For #86 I'll create a new branch and add the commits to that one, then PR to the upstream branch from there.

qcoptions.py
```diff
@@ -8,7 +8,7 @@
                   help="Minimum value of scan parameter", metavar="scanmin", default=0)
 parser.add_option("--scanmax", type="int", dest="scanmax",
                   help="Maximum value of scan parameter", metavar="scanmax", default=254)
-parser.add_option("--vfatmask", type="int", dest="vfatmask", 
-                  help="VFATs to be masked in scan & analysis applications (e.g. 0xFFFFF masks all VFATs)", metavar="vfatmask", default=0x0)
+parser.add_option("--vfatmask", type="int", dest="vfatmask",
+		              help="VFATs to be masked in scan & analysis applications (e.g. 0xFFFFF masks all VFATs)", metavar="vfatmask", default=0x0)
```
bdorney: are these tab characters (e.g. \t) here?  If so please change your editor such that when pressing "tab" a fixed number of spaces are inserted.  If you use vim this can be accomplished by editing:

`~/.vimrc`

And placing the following lines:

`
" show existing tab with 4 spaces width
set tabstop=4
" " when indenting with '>', use 4 spaces width
set shiftwidth=4
" " On pressing tab, insert 4 spaces
set expandtab
`

Into the file.

Otherwise this is fine for me.

qcoptions.py
```diff
@@ -9,6 +9,6 @@
 parser.add_option("--scanmax", type="int", dest="scanmax",
                   help="Maximum value of scan parameter", metavar="scanmax", default=254)
 parser.add_option("--vfatmask", type="int", dest="vfatmask",
-		              help="VFATs to be masked in scan & analysis applications (e.g. 0xFFFFF masks all VFATs)", metavar="vfatmask", default=0x0)
+                  help="VFATs to be masked in scan & analysis applications (e.g. 0xFFFFF masks all VFATs)", metavar="vfatmask", default=0x0)
```
bdorney: There is still a tab character here, e.g.  "\t"

bdorney (changes requested):

EklavyaFCB:
Had to remove an extra space elsewhere (in MSPL) for git to detect the tab removal for --vfatmask.

`cat -T qcoptions.py` showed no tab character.

bdorney (commented):
Can you confirm that this PR is ahead of the master branch? e.g. you do not have unmerged changes between your branch and the central remote's master branch?

EklavyaFCB:
My branch is up to date with the central repo:

`From github.com:cms-gem-daq-project/vfatqc-python-scripts`
`*branch            master     -> FETCH_HEAD`
`Already up-to-date.`

bdorney (approved):

Issue #104, opened 2017-07-19
Migrate Analysis tools to gem-plotting-tools


bdorney:
Based on discussion in July 18th GEM Operations & Online SW meeting it was decided to migrate all analysis tools from vfatqc-python-scripts to gem-plotting-tools

Issue #103, opened 2017-07-18
Extra parameters for latency scan


mexanick:
- need to propagate upstream to run_scans.py options to use the amc13 local trigger:

    - `--amc13local`
    - `--t3trig`
    - `--randoms`

- need to add the option to set trigger throttle

bdorney:
I think we can close this issue now as the requested features were implemented.

PR #105, opened 2017-07-19
Migration of Analysis Tools to gem-plotting-tools

bdorney:
Addressing:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/104

Tested:

http://cmsonline.cern.ch/cms-elog/998496

User should have following `$SHELL` variables declared:

`$BUILD_HOME`
`$DATA_PATH`

Then execute:

`setup/paths.sh`

Now also addressing:
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/103

Could we have review from @cbravo135 and @mexanick?


ultraScurve.py
```diff
@@ -14,8 +14,6 @@
                   help="Specify Output Filename", metavar="filename")
 parser.add_option("--latency", type="int", dest = "latency", default = 37,
                   help="Specify Latency", metavar="latency")
-parser.add_option("--mspl", type="int", dest = "MSPL", default = 4,
```
mexanick, line 17: why mspl option is removed?

ultraLatency.py
```diff
@@ -16,25 +16,25 @@
...
 parser.add_option("--randoms", type="int", default=0, dest="randoms",
                   help="Set up for using AMC13 local trigger generator to generate random triggers with rate specified",
                   metavar="randoms")
 parser.add_option("--t3trig", action="store_true", dest="t3trig",
                   help="Set up for using AMC13 T3 trigger input", metavar="t3trig")
-parser.add_option("--fakeTTC", action="store_true", dest="fakeTTC",
-                  help="Set up for using AMC13 local TTC generator", metavar="fakeTTC")
+parser.add_option("--throttle", type="int", default=0, dest="throttle",
+                  help="factor by which to throttle the input L1A rate, e.g. new trig rate = L1A rate / throttle", metavar="throttle")
+parser.add_option("--vt1", type="int", dest="vt1",
```
mexanick, line 34: do we really want to pass vt1 and vt2 here? Are they used in latency scan at all?

run_scans.py
```diff
@@ -104,8 +105,19 @@ def launchTestsArgs(tool, slot, link, chamber, scanmin, scanmax, nevts,
     cmd.append( "--scanmin=%d"%(scanmin) )
     cmd.append( "--scanmax=%d"%(scanmax) )
     cmd.append( "--nevts=%d"%(nevts) )
+    cmd.append( "--throttle=%i"%(throttle) )
```
mexanick, line 108: should be made optional, like amc13local below

mexanick (dismissed):

chamberInfo.py
```diff
@@ -1,5 +1,6 @@
 
 chamber_config = {
+    #0:"GE11-VI-L-CERN-0001"
```
cbravo135: please remove this line

chamberInfo.py
```diff
@@ -13,6 +14,7 @@
     }
 
 GEBtype = {
+    #0:"long"
```
cbravo135: please remove this line

fastLatency.py
```diff
@@ -15,8 +15,6 @@
                   help="Specify Output Filename", metavar="filename")
 parser.add_option("--vt1", type="int", dest="vt1",
                   help="VThreshold1 DAC value for all VFATs", metavar="vt1", default=100)
-parser.add_option("--mspl", type="int", dest = "MSPL", default = 4,
```
cbravo135, line 18: Did you move mspl to qcoptions as a default option?

pythonScript.py
```diff
@@ -12,6 +12,16 @@
 import sys, os, random, time
 
 if __name__ == "__main__":
+    print "Script depreciated, use instead:"
```
cbravo135: LOL, this script is here for the sole purpose of having a place to hack and try shit. I don't really care what goes in here.

qcoptions.py
```diff
@@ -1,30 +1,14 @@
...
-                  help="Tests to run, default is all", metavar="tests")
-parser.add_option("--doLatency", action="store_true", dest="doLatency",
-                  metavar="doLatency",
-                  help="[OPTIONAL] Run latency scan to determine the latency value")
-parser.add_option("--QC3test", action="store_true", dest="doQC3",
-                  metavar="doQC3",
-                  help="[OPTIONAL] Run a shortened test after covers have been applied")
-parser.add_option("--ztrim", type="float", dest="ztrim", default=0.0,
-                  help="Specify the p value of the trim", metavar="ztrim")
+parser.add_option("--mspl", type="int", dest = "MSPL", default = 4,
```
cbravo135: I think the default should be 1

run_scans.py
cbravo135: yeah, by default we shouldn't be touching this register for safety.

run_scans.py
```diff
@@ -134,22 +146,29 @@ def launchTestsArgs(tool, slot, link, chamber, scanmin, scanmax, nevts,
 
   from qcoptions import parser
 
-  parser.add_option("--series", action="store_true", dest="series",
```
cbravo135, line 137: why are you removing this option? we should be able to run the scans if series if we want...

run_scans.py
```diff
@@ -175,13 +196,32 @@ def launchTestsArgs(tool, slot, link, chamber, scanmin, scanmax, nevts,
...
                          [options.ztrim   for x in range(len(chamber_config))],
                          [options.config  for x in range(len(chamber_config))],
+                         [options.amc13local  for x in range(len(chamber_config))],
+                         [options.t3trig  for x in range(len(chamber_config))],
+                         [options.randoms for x in range(len(chamber_config))],
+                         [options.throttle for x in range(len(chamber_config))],
                          )
-
+            )
   if options.series:
```
cbravo135, line 205: removing the option above could make it crash here...

ultraLatency.py
cbravo135: Keep it, we have code below which will set all VT registers to the same value. There are situations where one might want to do that.

ultraLatency.py
```diff
@@ -198,6 +198,7 @@
         oh.setTriggerSource(ohboard,options.gtx,0x5) # GBT, 0x0 for GTX
         pass
 
+    oh.setTriggerThrottle(ohboard, options.gtx, options.throttle)
```
cbravo135, line 201: maybe we should only write this if it is different from 1, which should be the default value. Just makes me feel better if we are not always writing this register.

cbravo135 (changes requested):
Huge PR. Mostly fine, I put a few requests in the comments.

ultraScurve.py
bdorney: All of the scan scripts had this option hard coded in each of the scan options.  So I moved it to qcoptions.py

See lines 3-4 of:

https://github.com/bdorney/vfatqc-python-scripts/blob/develop/qcoptions.py

run_scans.py
bdorney: The default value that is passed is 0, and the default value in the options is 0.  I understood based on the implementation a value of 0 means no throttle.

So there should be no problem here. No?

chamberInfo.py
bdorney: It makes life easier on QC8, I'd like to keep it.

chamberInfo.py
bdorney: It makes life easier on QC8, I'd like to keep it.

fastLatency.py
bdorney: Yes, as I responded above:

> All of the scan scripts had this option hard coded in each of the scan options. So I moved it to qcoptions.py
> See lines 3-4 of:
> https://github.com/bdorney/vfatqc-python-scripts/blob/develop/qcoptions.py

pythonScript.py
bdorney: Scripts that are purely used for testing new implementations should not be kept in the repo.  You should have ignored by your gitignore.

Deleting.

qcoptions.py
bdorney: Sure, changing.

run_scans.py
bdorney: It's not removed, I alphabetised the options menu, see line 158:

https://github.com/bdorney/vfatqc-python-scripts/blob/develop/run_scans.py

run_scans.py
bdorney: Again:

> t's not removed, I alphabetised the options menu, see line 158:
> https://github.com/bdorney/vfatqc-python-scripts/blob/develop/run_scans.py

ultraLatency.py
bdorney: I thought register value of 0 means no throttle.

Additionally why does it matter if we write to this register by default, from power on it should be set to not throttle, we're just writing this same value back into it no?

I don't see the problem.

run_scans.py
mexanick: Ok, it makes sense to leave it as is. However in general we try to minimize reg access ops, i.e. do not execute what is unneccessary

ultraLatency.py
mexanick: Default is 0. In principle i agree with Cameron, but it might be beneficial to always write the default so we don,t run with unexpected leftover value

mexanick (approved):
I'm fine with this PR. Please finalize requests from Cameron and we can merge

ultraLatency.py
bdorney: Yeah I think it's better to write it at the start of the scan, that way you know what it's set too.

chamberInfo.py
cbravo135: Can you add one for the coffin setup too then?

ultraLatency.py
cbravo135: To me this is just asking for it to make a mistake a write an incorrect value. We know the default setting is more likely to work than a write, might be (probably) negligible but we do not know what our miswrite rate is yet. This is probably fine. So the throttle is systematically what a counter must reach before the next L1A is distributed?

bdorney: On the contrary, how do you know what this register is set to when you launch another scan?  Say you launch a scan with throttle 500, then you launch another scan with throttle 0.  In between scans the throttle is never overwritten so your next scan launched with throttle 0 is run with throttle 500.

confChamber right now is not changing this method (maybe it should, but I would prefer not since it seems to only be relevant for latency scans).

I think we should follow the statement of Misha above.  Leave it as it is, and ensure it's written each time but make the default value assigned (0) correspond to the unthrottled case.

If we notice this causes instabilities I would be happy then to submit a new issue on the repo and resolve the bug fix but I think this implementation is preferred.  You know the state of the system when you start.

chamberInfo.py
bdorney: Maybe it's better to make a symlink and have three separate chamberInfo.py files for P5, coffin, and QC8?  This mirrors the functionality in cmsgemos with the connections file.  

Although this might increase code maintenance if we expand the dictionaries stored in chamberInfo.py, e.g. what Jared is suggesting in:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/86

ultraLatency.py
cbravo135: Yeah, the only way to fix this would be with a read. I agree, leave it as is.

cbravo135 (approved):

Issue #107, opened 2017-07-24
Feature Request: Add "--internal" to run_scans.py


bdorney:
Allow run_scans.py to use the internal trigger when running a latency scan.

jsturdy:
What is the purpose for this?

bdorney:
Software testing and debugging...requested by DOC

jsturdy:
I ask because this option is definitely a debug option, and I don't see a strong use case to run it on all chambers simultaneously, so could be kept out of `run_scans` for simplicity

Issue #106, opened 2017-07-24
Issue: run_scans.py does not pass options.nevt to every tool


bdorney:
DOC reports `run_scans.py` does not pass argument given to `--nevt` to `tool=trimChamber.py`

Looks like `launchTestsArgs()` has the nevt argument but is not passing it to every tool.

We should think about a review of `launchTestsArgs()` to make sure that the set of all common parameters are being passed to the appropriate tools when launching scans.

In the short term propose to change:

`  cmd = ["%s"%(tool),"-s%d"%(slot),"-g%d"%(link),"--shelf=%i"%(shelf)]`

to:

`  cmd = ["%s"%(tool),"-s%d"%(slot),"-g%d"%(link),"--shelf=%i"%(shelf), "--nevts=%i"%(nevts)]`

Looks `options.nevts` is called by each of the following scans:
> ultraLatency.py
> ultraScurve.py
> ultraThreshold.py

Additionally trimChamber.py is **not** using this option in each of its calls to ultraScurve.py  Suggest to modify this script accordingly.

PR #108, opened 2017-07-24
Adding more cmd line options to run_scans.py

bdorney:
Addressing:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/107
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/106
https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/112

@cbravo135 and @mexanick 

Additionally added a new dictionary in chamberInfo.py:

`chamber_vfatMask`

This will now allow run_scans.py to send vfatmask parameter correctly to each chamber.  The defaults assigned for P5 are 0x0 for all links, 0x0 for coffin setup, and 0xF40400 for the cosmic stand.

Now also requires:
https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/13


trimChamber.py
```diff
@@ -86,8 +86,10 @@
             "-s%d"%(options.slot),
             "-g%d"%(options.gtx),
             "--filename=%s"%(filename0),
-	        "--vfatmask=%i"%(options.vfatmask)]
-           )
+            "--vfatmask=%i"%(options.vfatmask),
```
jsturdy: This is definitely the wrong way to do this, since there is no expectation that the `vfatmask` will be the same for all chambers under consideration, probably best would be to have it stored in the `chamber_info` file per chamber

bdorney: Actually trimChamber.py seems to be the place to put this, from what I understood of your comment though run_scans.py should read in a dict from chamber_info and give the vfatmask option to the chamber of interest when called as it does per the other values.

I will update run_scans.py to have this functionality.

bdorney: Functionality now included in chamber_vfatMask dictionary and used by run_scans.py

run_scans.py
```diff
@@ -21,7 +22,7 @@ def launchTestsArgs(tool, shelf, slot, link, chamber, scanmin, scanmax, nevts,
   #Build Commands
   setupCmds = []
   preCmd = None
-  cmd = ["%s"%(tool),"-s%d"%(slot),"-g%d"%(link),"--shelf=%i"%(shelf)]
+  cmd = ["%s"%(tool),"-s%d"%(slot),"-g%d"%(link),"--shelf=%i"%(shelf), "--nevts=%i"%(nevts), "--vfatmask=%s"%(vfatmask)]
```
jsturdy: you've tested this?  some places you're passing it as an `int` (`%i`), some places as a `string` (`%s`)
I'd have probably naively gone with `hex` (`%x` or `0x%x`)

bdorney: Mmhmm...an oversight for sure.  I'll try to incorporate the suggesting and then re-test.

trimChamber.py
```diff
@@ -232,7 +232,7 @@
             "-s%d"%(options.slot),
             "-g%d"%(options.gtx),
             "--filename=%s"%(filenameFinal),
-	        "--vfatmask=%i"%(options.vfatmask),
+	        "--vfatmask=%x"%(options.vfatmask),
```
jsturdy: check tabs here (L235)

run_scans.py
```diff
@@ -22,7 +22,7 @@ def launchTestsArgs(tool, shelf, slot, link, chamber, vfatmask, scanmin, scanmax
   #Build Commands
   setupCmds = []
   preCmd = None
-  cmd = ["%s"%(tool),"-s%d"%(slot),"-g%d"%(link),"--shelf=%i"%(shelf), "--nevts=%i"%(nevts), "--vfatmask=%s"%(vfatmask)]
+  cmd = ["%s"%(tool),"-s%i"%(slot),"-g%i"%(link),"--shelf=%i"%(shelf), "--nevts=%i"%(nevts), "--vfatmask=%i"%(vfatmask)]
```
jsturdy: probably doesn't affect operation, but for consistency, change L25 to `--vfatmask=0x%x` (unless while testing this was found to not work)

qcoptions.py
```diff
@@ -8,7 +8,7 @@
                   help="Minimum value of scan parameter", metavar="scanmin", default=0)
 parser.add_option("--scanmax", type="int", dest="scanmax",
                   help="Maximum value of scan parameter", metavar="scanmax", default=254)
-parser.add_option("--vfatmask", type="int", dest="vfatmask",
+parser.add_option("--vfatmask", type="int", dest="vfatmask",
```
jsturdy: And since I'm making other change requests, prefer to remove all trailing whitespace (L11)

jsturdy (changes requested):

run_scans.py
bdorney: I had tried `--vfatmask=%x` and it didn't work (learned why it should have been `--vfatmask=0x%x`) after googling more about optparser class:

https://docs.python.org/2/library/optparse.html#optparse-standard-option-types

Relying on optparser to identify the type here also allows the user to enter both hex or a binary number (whereas the `--vfatmask=0x%x` does not) which did work in testing.  Since optparser automatically handles this I don't think the change is necessary.

trimChamber.py
bdorney: Mmhmm, not sure how that slipped through

run_scans.py
jsturdy: OK, but the point here was that the user is not going to be specifying anything in this code (the value is passed in from somewhere else), so what is printed to the screen and logfile, should be as readable as possible (decimal representation is the least readable and binary is long).

bdorney: Okay, understood.  I will change to:

`"--vfatmask=0x%x"`

And make the appropriate modifications in chamberInfo.py

But in this case the log file will *still* show the number as decimal in some instances since optparser itself does the conversion. e.g. using the `--debug` option with `run_scans.py` will have a hex argument:

0xF40400

Displayed as:

15991808

Due to the conversion by optparser.

qcoptions.py
jsturdy: Still trailing whitespace (since github told me my comment was outdated)

run_scans.py
jsturdy: Which line makes this print out?  We should just fix that.
Also, what needs to change in `chamberInfo.py`?  The version in this PR looked fine just a minute ago

qcoptions.py
bdorney: Mmhmm, maybe It's something in my .vimrc then...from terminal in my editor I do not see these extra spaces but they do appear:

https://github.com/bdorney/vfatqc-python-scripts/blob/master/qcoptions.py

If I highlight the text in github...

run_scans.py
bdorney: Looks like:

https://github.com/bdorney/vfatqc-python-scripts/blob/master/run_scans.py#L198

Maybe calling `hex()` here will ensure the log file has this written in hex format.

For chamberInfo.py if we have:

`--vfatmask=0x%x`

I understood from:

https://docs.python.org/2/library/optparse.html#optparse-standard-option-types

That the number should be formatted like "0x...." but from [chamber_vfatmask](https://github.com/bdorney/vfatqc-python-scripts/blob/bc120ab04e6b20eb302585deeaf6d2bda0bd4ddf/chamberInfo.py#L37)

The inputs are `0x0` so they should instead be stored as `0` otherwise we would pass `0x0x0` to vfatmask which in the `append()` command is a string so it is not interpreted as a hex number.

Basically we need to resolve 0x... vs. 0x0x... or have I misunderstood?

jsturdy: You've (possibly) misunderstood, once the number is extracted from the `chamberInfo.py` it is an `int`, so it's machine readable, then when you pass this to the line we're discussing, it will be correctly converted to `hex`.
Though I'd double check with some test, just to convince yourself

qcoptions.py
jsturdy: could be the way that github manages comments linked to code, as i had a similar issue just now where what i see on the PR dashboard and what I see in the latest version of the file in the PR aren't in agreement...
UPDATE nope, just checked and confirm what you've seen, could be that it's a single space, or a tab (not sure what `vim` is showing you?

bdorney: I added some statements to my ~/.vimrc to remove trailing white space on *.py files whenever the save action is performed by vim (e.g. `:w`).  Hopefully this should solve the problem...but git may not have realized anything in the file has changed.

jsturdy (approved):
ok, if it works, i sign off

bdorney:
Unit tests completed, tests on cosmic stand using `run_scans.py` successful:

http://cmsonline.cern.ch/cms-elog/1003922

Will merge once cms-gem-daq-project/gem-plotting-tools#13 is approved (since this PR is dependent)

bdorney:
https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/13

Has been merged.  Now merging this request

Issue #109, opened 2017-08-09
setup scripts do naughty things


jsturdy:
Sourcing the `setup/paths.sh` script leaves the user in a specific directory, which is not good practice, especially when used in a profile init script
Probably everything below [#L44](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/master/setup/paths.sh#L44) should not do a `cd`

PR #110, opened 2017-08-09
Don't use cd in setup/path.sh

lmoureaux:
Addresses bug #109 assuming that $BUILD_HOME is an absolute path.

### DIFF
--- a/setup/paths.sh
+++ b/setup/paths.sh
@@ -23,18 +23,14 @@ fi
 if [[ -n "$GEM_PYTHON_PATH" ]]; then
     echo GEM_PYTHON_PATH $GEM_PYTHON_PATH
 else
-    echo "GEM_PYTHON_PATH not set"
-    echo "Setting up GEM_PYTHON_PATH"
-    source $BUILD_HOME/cmsgemos/setup/paths.sh
+    echo "GEM_PYTHON_PATH not set, please source \$BUILD_HOME/cmsgemos/setup/paths.sh"
 fi
 
 # Checking GEM_PLOTTING_PROJECT
 if [[ -n "$GEM_PLOTTING_PROJECT" ]]; then
     echo GEM_PLOTTING_PROJECT $GEM_PLOTTING_PROJECT
 else
-    echo "GEM_PLOTTING_PROJECT not set"
-    echo "Setting up GEM_PLOTTING_PROJECT"
-    source $BUILD_HOME/gem-plotting-tools/setup/paths.sh
+    echo "GEM_PLOTTING_PROJECT not set, please source \$BUILD_HOME/gem-plotting-tools/setup/paths.sh"
 fi
 
 # Adding Scan Applications to Path
--- a/setup/paths.sh
+++ b/setup/paths.sh
@@ -38,18 +38,14 @@ else
 fi
 
 # Adding Scan Applications to Path
-#cd $BUILD_HOME/cmsgemos/gempython/tools
-#export PYTHONPATH=$PYTHONPATH:$PWD
+#export "PYTHONPATH=$PYTHONPATH:$BUILD_HOME/cmsgemos/gempython/tools"
 
-cd $BUILD_HOME/vfatqc-python-scripts
-export PATH=$PATH:$PWD
-export PYTHONPATH=$PYTHONPATH:$PWD
+export "PATH=$PATH:$BUILD_HOME/vfatqc-python-scripts"
+export "PYTHONPATH=$PYTHONPATH:$BUILD_HOME/vfatqc-python-scripts"
 
-cd $BUILD_HOME/vfatqc-python-scripts/setup
-export PATH=$PATH:$PWD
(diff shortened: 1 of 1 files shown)


setup/paths.sh
```diff
@@ -38,18 +38,14 @@ else
...
+export "PYTHONPATH=$PYTHONPATH:$BUILD_HOME/vfatqc-python-scripts"
 
-cd $BUILD_HOME/vfatqc-python-scripts/setup
-export PATH=$PATH:$PWD
+export "PATH=$PATH:$BUILD_HOME/vfatqc-python-scripts/setup"
 
 # Done
-cd $BUILD_HOME/vfatqc-python-scripts
-export GEM_VFATQC_PROJECT=$PWD
+export "GEM_VFATQC_PROJECT=$BUILD_HOME/vfatqc-python-scripts"
```
cbravo135: Isn't it cleaner to do GEM_VFATQC_PROJECT first and then use it to make the others. This all could be done in 3 simpler lines.

jsturdy: Yeah, probably.
Also what is `GEM_VFATQC_PROJECT` actually used for?

jsturdy (changes requested):

setup/paths.sh
```diff
@@ -38,18 +38,14 @@ else
 fi
 
 # Adding Scan Applications to Path
-#cd $BUILD_HOME/cmsgemos/gempython/tools
-#export PYTHONPATH=$PYTHONPATH:$PWD
+#export "PYTHONPATH=$PYTHONPATH:$BUILD_HOME/cmsgemos/gempython/tools"
```
bdorney: shouldn't this be uncommented?

bdorney (changes requested):
Agree w/@cbravo135 regarding cleaning up setup scripts.  Probably the same could be done in gem-plotting-tools

setup/paths.sh
lmoureaux: I didn't find any use of `GEM_VFATQC_PROJECT` in `vfatqc_python_tools`, `cmsgemos` and `gem-plotting-tools`

setup/paths.sh
lmoureaux: I didn't want to introduce any breaking change.

jsturdy: * it was commented in the previous version (according to the diff in this PR)
* also no, the setup script in `cmsgemos` should appropriately set the `PYTHONPATH` for use with the `cmsgemos` code, if it doesn't (it [does)](https://github.com/cms-gem-daq-project/cmsgemos/blob/generic-amc-release-v3/setup/paths.sh#L67) then open an issue there

setup/paths.sh
jsturdy: In that case, then you can make it a script local variable (if even at all), i.e., don't `export` it

bdorney (approved):

setup/paths.sh
```diff
@@ -38,18 +38,12 @@ else
...
 
-cd $BUILD_HOME/vfatqc-python-scripts
-export PATH=$PATH:$PWD
-export PYTHONPATH=$PYTHONPATH:$PWD
-
-cd $BUILD_HOME/vfatqc-python-scripts/setup
-export PATH=$PATH:$PWD
+export "PATH=$PATH:$BUILD_HOME/vfatqc-python-scripts"
+export "PATH=$PATH:$BUILD_HOME/vfatqc-python-scripts/setup"
```
jsturdy: Line 44 should be removed unless there is a good reason to keep it

setup/paths.sh
```diff
@@ -38,18 +38,12 @@ else
...
+export "PYTHONPATH=$PYTHONPATH:$BUILD_HOME/vfatqc-python-scripts"
 
 # Done
-cd $BUILD_HOME/vfatqc-python-scripts
-export GEM_VFATQC_PROJECT=$PWD
 echo GEM_VFATQC_PROJECT $GEM_VFATQC_PROJECT
```
jsturdy: As per comments, if `GEM_VFATQC_PROJECT` isn't used outside this script, don't `export` it and remove line 48

setup/paths.sh
jsturdy: As is, just remove line 41

jsturdy (changes requested):
* Prefer sandboxing of all setup scripts
 * Change  https://github.com/lmoureaux/vfatqc-python-scripts/blob/d97e2a0ae35ad541e2d48730271bcd24a2a90fe5/setup/paths.sh#L22-L29 to instruct the user to ensure that `
$BUILD_HOME/cmsgemos/setup/paths.sh` gets sourced
 * Same for https://github.com/lmoureaux/vfatqc-python-scripts/blob/d97e2a0ae35ad541e2d48730271bcd24a2a90fe5/setup/paths.sh#L31-L38
* Don't arbitrarily add locations to the `$PATH`, this will have negative consequences

jsturdy (approved):

jsturdy (approved):
Closes #109

PR #111, opened 2017-08-09
Dictionary values config of VFATs

EklavyaFCB:
Addressing:

#86

Added `chamber_vfatDACSettings` dictionary in `ChamberInfo.py`, such as :

```chamber_vfatDACSettings = {
    #Cosmic Stand
    0: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,}
    } 
```

Note: the above shows only the relevant bit for the cosmic stand with the g0 link. Same lines have been added for the P5 and COFFIN as well. To use adequately one should make sure the other irrelevant parts are duly commented-out.

 Updated `confChamber.py` by:

Adding an if-structure that would use the `writeAllVFATs` function to overwrite the default values (written by the `biasAllVFATs` method) if the link provided as an argument exists in the `chamber_vfatDACSettings` dictionary, and reconfigure the chamber’s IPreampIn, IPreampFeed, IPreampOut, IShaper, IShaperFeed and IComp values.

Usage example (for cosmic stand):

`confAllChambers.py --shelf=3 -s2 -g0 --config`

`confChamber.py -s2 -g0 --shelf=3`

Changes the VFAT parameters are expected. Can be seen using `vfat_info_uhal.py -s2 -g0 --shelf=3` (for cosmic stand).

Tested and confirmed to work:
 
http://cmsonline.cern.ch/cms-elog/1003518

--

Older versions of tests can be found at:

http://cmsonline.cern.ch/cms-elog/999820
http://cmsonline.cern.ch/cms-elog/999477

(Note: these last 2 e-logs above used a --dictConfig or --ccc flag to enable dictionary values. Since then, the code has been updated to automatically use the dictionary values if the provided link exists in the dictionary, therefore rendering the flags and the shown usage in logs outdated).

### DIFF
--- a/chamberInfo.py
+++ b/chamberInfo.py
@@ -34,7 +34,7 @@ GEBtype = {
     9:"long"
     }
 
-chamber_defaults = {
+chamber_vfatDACSettings = {
     #Coffin Setup
     #0: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
     #Cosmic Stand
--- a/chamberInfo.py
+++ b/chamberInfo.py
@@ -33,3 +33,21 @@ GEBtype = {
     8:"long",
     9:"long"
     }
+
+chamber_defaults = {
+    #Coffin Setup
+    #0: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    #Cosmic Stand
+    #0: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    #Point 5
+    0: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    1: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    2: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    3: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    4: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    5: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    6: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    7: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    8: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,},
+    9: {"IPreampIn": 168, "IPreampFeed": 80, "IPreampOut": 150, "IShaper": 150, "IShaperFeed": 100, "IComp": 90,}
+    }
--- a/confAllChambers.py
+++ b/confAllChambers.py
@@ -3,9 +3,8 @@
 def launch(args):
   return launchArgs(*args)
 
-def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,dictConfig,cName,ztrim):
+def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,cName,ztrim):
     import datetime,os,sys
-    import subprocess
     from subprocess import CalledProcessError
     from chamberInfo import chamber_config
     from gempython.utils.wrappers import runCommand
@@ -21,12 +20,7 @@ def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,dictConfig,cName,ztrim):
 
     if config:
         cmd.append("--vt1bump=%d"%(vt1bump))
-        if options.dictConfig:
-            cmd.append("--dictConfig")
-            pass
-        else:
-            cmd.append("--vfatConfig=%s/configs/z%.1f/vfatConfig_%s.txt"%(dataPath,ztrim,cName))
-            pass
+        cmd.append("--vfatConfig=%s/configs/z%.1f/vfatConfig_%s.txt"%(dataPath,ztrim,cName))
         cmd.append("--chConfig=%s/configs/z%.1f/chConfig_%s.txt"%(dataPath,ztrim,cName))
         pass
     else:
@@ -58,8 +52,6 @@ if __name__ == '__main__':
 
     parser.add_option("--config", action="store_true", dest="config",
                       help="Set Configuration from simple txt files", metavar="config")
-    parser.add_option("--dictConfig", action="store_true", dest="dictConfig", default=False,
-                      help="Configure VFATs to custom chamber_default values", metavar="dictConfig")
     parser.add_option("--run", action="store_true", dest="run",
                       help="Set VFATs to run mode", metavar="run")
     parser.add_option("--series", action="store_true", dest="series",
@@ -83,7 +75,6 @@ if __name__ == '__main__':
                         [options.vt1bump   for x in range(len(chamber_config))],
                         [options.config    for x in range(len(chamber_config))],
                         [chamber_config[x] for x in chamber_config.keys()],
-                        [options.dictConfig for x in range(len(chamber_config))],
                         [options.ztrim     for x in range(len(chamber_config))],
                   )
             )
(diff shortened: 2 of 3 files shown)


EklavyaFCB:
@bdorney Forgot to tag you. 

--

Additionally, when I did git pull from the central repo:

`remote: Counting objects: 1, done.`
`remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0`
`Unpacking objects: 100% (1/1), done.`
`From github.com:cms-gem-daq-project/vfatqc-python-scripts`
`*branch            master     -> FETCH_HEAD`
`Already uptodate!`
`Merge made by recursive.`

Branch is up to date -- no unmerged changes currently.

confAllChambers.py
```diff
@@ -3,7 +3,7 @@
 def launch(args):
   return launchArgs(*args)
 
-def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,cName,ztrim):
+def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,dictConfig,cName,ztrim):
     import datetime,os,sys
     import subprocess
     from subprocess import CalledProcessError
```
bdorney: Looks like there is a redundant import here, if you're only using `CalledProcessError` you should not have line 8, if you're using more from subprocess I don't think you need Line 8 & 9.

confAllChambers.py
```diff
@@ -21,8 +21,14 @@ def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,cName,ztrim):
 
     if config:
         cmd.append("--vt1bump=%d"%(vt1bump))
-        cmd.append("--vfatConfig=%s/configs/z%.1f/vfatConfig_%s.txt"%(dataPath,ztrim,cName))
+        if options.dictConfig:
+            cmd.append("--dictConfig")
+            pass
+        else:
+            cmd.append("--vfatConfig=%s/configs/z%.1f/vfatConfig_%s.txt"%(dataPath,ztrim,cName))
```
bdorney: Lines 24-28, these should not be mutually exclusive.  In `chamber_defaults` you provide register values to setup the preamp & shaper of the VFAT2.  The `vfatConfig.txt` provides information on the threshold and trim range registers.  The _both_ sets information of information are required to properly setup a detector.

Additionally vfatConfig provides unique settings for **each** VFAT while chamber_defaults is applied to **all** VFATs.

This needs to be re-implemented.  e.g. if `--config` is supplied it should *always* append the `--vfatconfig` option.

confAllChambers.py
```diff
@@ -52,6 +58,8 @@ def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,cName,ztrim):
 
     parser.add_option("--config", action="store_true", dest="config",
                       help="Set Configuration from simple txt files", metavar="config")
+    parser.add_option("--dictConfig", action="store_true", dest="dictConfig", default=False,
+                      help="Configure VFATs to custom chamber_default values", metavar="dictConfig")
```
bdorney: To avoid option bloat here and argument bloat in launchArgs remove entirely the `--dictConfig` option _and_ argument.  

Instead a more elegant solution the code (probably inside launchArgs) should check if the link in question is present in chamber_defaults, if it is then the settings provided in chamber_defaults should be used.  If not the settings provided in vfat_user_functions.py L15:

https://github.com/cms-gem-daq-project/cmsgemos/blob/07d68dc291aac5f51f9e5d12fb9e3bd38c16cac7/gempython/tools/vfat_user_functions_uhal.py#L15

Should be used instead.

For e like:

```
if link not in chamber_defaults.keys():
    Defaults from vfat_user_functions.py line 15 applied
    pass
else:
    Settings stored in chamber_defaults applied
    pass
```

confAllChambers.py
```diff
@@ -75,6 +83,7 @@ def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,cName,ztrim):
                         [options.vt1bump   for x in range(len(chamber_config))],
                         [options.config    for x in range(len(chamber_config))],
                         [chamber_config[x] for x in chamber_config.keys()],
+                        [options.dictConfig for x in range(len(chamber_config))],
```
bdorney: re-design as described.

confAllChambers.py
```diff
@@ -83,7 +92,7 @@ def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,cName,ztrim):
         print "Configuring chambers in serial mode"
         for link in chamber_config.keys():
             chamber = chamber_config[link]
-            launchArgs(options.shelf,link,options.slot,options.run,options.vt1,options.vt1bump,options.config,chamber,options.ztrim)
+            launchArgs(options.shelf,link,options.slot,options.run,options.vt1,options.vt1bump,options.config,options.dictConfig,chamber,options.ztrim)
```
bdorney: redesign as described

confAllChambers.py
```diff
@@ -102,6 +111,7 @@ def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,cName,ztrim):
                                                 [options.vt1bump   for x in range(len(chamber_config))],
                                                 [options.config    for x in range(len(chamber_config))],
                                                 [chamber_config[x] for x in chamber_config.keys()],
+                                                [options.dictConfig  for x in range(len(chamber_config))],
```
bdorney: redesign as described

chamberInfo.py
```diff
@@ -33,3 +33,21 @@
     8:"long",
     9:"long"
     }
+
+chamber_defaults = {
```
bdorney: Let's rename this to a more accurate description.

You're storing parameters for the VFAT preamp, shaper, and comparator, these are applied to each VFAT on a detector.  This is more for the VFAT than the "detector."

Also the defaults are provided at:

https://github.com/cms-gem-daq-project/cmsgemos/blob/generic-amc-release-v3/gempython/tools/vfat_user_functions_uhal.py#L15

Instead let's call this dict **chamber_vfatDACSettings** instead

confChamber.py
```diff
@@ -6,22 +6,23 @@
...
 from gempython.tools.vfat_user_functions_uhal import *
-
+from chamberInfo import chamber_defaults
 from qcoptions import parser
 
-parser.add_option("--filename", type="string", dest="filename", default=None,
-                  help="Specify file containing settings information", metavar="filename")
 parser.add_option("--chConfig", type="string", dest="chConfig", default=None,
                   help="Specify file containing channel settings from anaUltraSCurve", metavar="chConfig")
+parser.add_option("--dictConfig", action="store_true", dest="dictConfig", default=False,
```
bdorney: same comment as above regarding `--dictConfig` option

confChamber.py
```diff
@@ -87,5 +88,20 @@
     except Exception as e:
         print '%s does not seem to exist'%options.filename
         print e
-print 'Chamber Configured'
 
+if options.dictConfig:
+    try:
+        print "Configuring VFATs with dictionary values"
+
+        for x in range (len(chamber_defaults)):
```
bdorney: This is a mistake, confChamber.py is called once per detector either by the user individually or by confAllChambers.py.  Each execution corresponds to one detector.

Here you're looping over all the elements of the chamber_defaults dictionary.  So you'll be writing these registers `len(chamber_defaults)` times and the register values that are stored will correspond to the last element `len(chamber_defaults)-1` rather than the link given by the `-g` option.

Please revise, also taking into account the suggested redesign above (e.g. if a link isn't present in the dictionary it should default to those values stored in vfat_user_functions.py)

bdorney (changes requested):
There are several mistakes in the code proposed here.  Additionally I'd like this implementation to be redesigned such that if a link isn't mentioned in the chamber_vfatDACSettings (renamed from chamber_defaults) dict then the default settings shown in vfat_user_functions.py should be used (as they are now).

confAllChambers.py
EklavyaFCB: Q: 

Aren't the default values from `vfat_user_functions.py` l.15 that you linked applied every time when the `biasAllVFATs(ohboard,options.gtx,0x0,enable=False)` method is called at l.43 of `confChamber.py` ? If so, then shouldn't we just require:

```
if link in chamber_defaults.keys():
    Settings stored in chamber_defaults applied
    pass
```

as the alternative would be re-writing the defaults that have already been written at l.43 ? Unless, I'm misunderstanding what `biasAllVFATs` does, but it seems to me it's writing the register values with the `writeAllVFATs` method using the `parameters.defaultValues` values.

bdorney: Yes the case structure you showed above is probably a better way to apply it; the goal here was just to provide an example.  It's more likely that `confChamber.py` is the correct place to put this case structure since this is performing the actual writing of registers; the job of `confAllChambers.py` is to ensure the call of `confChamber.py` for a given link is correct.

In any case, you get the idea.

bdorney:
Please update the description of this pull request to reflect the status of the code, for example the `--dictConfig` option no longer exists.

Also please point to the results of testing these updated changes.

EklavyaFCB:
Here is the relevant e-log: http://cmsonline.cern.ch/cms-elog/1003518

The dictConfig is the name of my git branch, do you want me to rename it ? It will mean deleting the old branch and pushing the new one I believe. Unless by description you meant the first comment ?

bdorney:
No I want you to edit the description of this pull request at the top of the page which explains this pull request and shows usage examples.  For instance you have written:

> Usage example (for cosmic stand):
>
> `confAllChambers.py --shelf=3 -s2 -g0 --config --dictConfig`
>
> `confChamber.py -s2 -g0 --shelf=3 --dictConfig`
>
> Changes the VFAT parameters are expected. Can be seen using vfat_info_uhal.py -s2 -g0 --shelf=3 (for cosmic stand).

Which will generate an error if the user attempts to execute because the `--dictConfig` option no longer exists.

EklavyaFCB:
Have updated.

confChamber.py
```diff
@@ -6,22 +6,21 @@
```
bdorney, line 7: After line 4 please add:

```
Modified by: Eklavya Sarkar eklavya.sarkar@cern.ch
```

bdorney (approved):

bdorney:
Extensive review and, more importantly, testing on cosmic stand show this PR behaves as expected.  Going ahead with the merge.

Issue #117, opened 2017-08-21
Bug Report: run_scans.py does not pass shelf argument when --config option is used


bdorney:
## Brief summary of issue
`run_scans.py` does not properly apply the `--shelf` argument to the `preCmd` call when the `--config` option is used

### Types of issue

- [X] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

When supplying the `--config` and `--shelf=X` arguments to `run_scans.py` the `--shelf` argument should be passed to the call of `confChamber.py` at Lines [37](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/4b03d64b08c1bc9b70ae2fcbd5f4ea624d289c6f/run_scans.py#L37), [45](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/4b03d64b08c1bc9b70ae2fcbd5f4ea624d289c6f/run_scans.py#L45), and [135](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/4b03d64b08c1bc9b70ae2fcbd5f4ea624d289c6f/run_scans.py#L135).

## Current Behavior

When using the `--config` and `--shelf` options, regardless of the value assigned to `--shelf` the call of `confChamber.py` executes the configuration using `options.shelf=1` since it is the default stored in `standardopts.py` at Line [8](https://github.com/cms-gem-daq-project/cmsgemos/blob/b85f5bb29772760ec22ea772cc02e63dd4623699/gempython/utils/standardopts.py#L8)

This was a partial reason which lead to bad data being recorded at:
http://cmsonline.cern.ch/cms-elog/1005281

### Steps to Reproduce (for bugs)

1. Execute `run_scans.py` and supply the `--config` and `--shelf=X` options for `shelf != 1`

## Possible Solution (for bugs)

When setting up the preCmd list insert the string:

```
"--shelf=%i"%(shelf)
```

as an element.

Additionally the current implementation may be a challenge to maintain.  Presently `preCmd` is defined at Line [24](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/4b03d64b08c1bc9b70ae2fcbd5f4ea624d289c6f/run_scans.py#L24) and set to `None`.  

[...]

PR #118, opened 2017-08-22
Addressing issue #117

bdorney:
## Description
Addressing https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/117.

`run_scans.py` will now properly pass the `--shelf` argument to the `preCmd` when the `--config` option is used.

### Types of changes

## Motivation and Context

See above.  The above issue would prevent scans from being taken with the correct configuration.

## How Has This Been Tested?

Tests performed here: http://cmsonline.cern.ch/cms-elog/1005139

And analyzed here: http://cmsonline.cern.ch/cms-elog/1005447

The `--shelf=3` option was properly passed with calling the `--config` option when using commit: 56476318321bb01d254094c9631cd69285659935.

summary_vt1_20
fitsummary_vt1_20

## Checklist:

### DIFF
--- a/run_scans.py
+++ b/run_scans.py
@@ -21,7 +21,7 @@ def launchTestsArgs(tool, shelf, slot, link, chamber, vfatmask, scanmin, scanmax
 
   #Build Commands
   setupCmds = []
-  preCmd = None 
+  preCmd = None
   cmd = ["%s"%(tool),"-s%i"%(slot),"-g%i"%(link),"--shelf=%i"%(shelf), "--nevts=%i"%(nevts), "--vfatmask=0x%x"%(vfatmask)]
   if tool == "ultraScurve.py":
     scanType = "scurve"
@@ -132,11 +132,9 @@ def launchTestsArgs(tool, shelf, slot, link, chamber, vfatmask, scanmin, scanmax
       pass
     log = file("%s/scanLog.log"%(dirPath),"w")
     if preCmd and config:
-      #runCommand(preCmd,log)
-      runCommand(preCmd)
+      runCommand(preCmd,log)
       pass
-    #runCommand(cmd,log)
-    runCommand(cmd)
+    runCommand(cmd,log)
   except CalledProcessError as e:
     print "Caught exception",e
     pass
@@ -197,7 +195,7 @@ if __name__ == '__main__':
                          [options.slot for x in range(len(chamber_config))],
                          chamber_config.keys(),
                          chamber_config.values(),
-                         chamber_vfatMask.values(),
+                         [hex(vfatmask) for vfatmask in chamber_vfatMask.values()],
                          [options.scanmin for x in range(len(chamber_config))],
                          [options.scanmax for x in range(len(chamber_config))], 
                          [options.nevts   for x in range(len(chamber_config))],
--- a/run_scans.py
+++ b/run_scans.py
@@ -21,7 +21,7 @@ def launchTestsArgs(tool, shelf, slot, link, chamber, vfatmask, scanmin, scanmax
 
   #Build Commands
   setupCmds = []
-  preCmd = None
+  preCmd = ["confChamber.py","-s%i"%(slot),"-g%i"%(link),"--shelf=%i"%(shelf)]
   cmd = ["%s"%(tool),"-s%i"%(slot),"-g%i"%(link),"--shelf=%i"%(shelf), "--nevts=%i"%(nevts), "--vfatmask=0x%x"%(vfatmask)]
   if tool == "ultraScurve.py":
     scanType = "scurve"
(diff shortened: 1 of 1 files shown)


bdorney:
The quality check:

> code-quality/landscape — Waiting for status to be reported

Seems to be stuck at this state since yesterday.  Any ideas as to why?

jsturdy:
Hmm, not sure... I see on the [dashboard](https://landscape.io/github/cms-gem-daq-project/vfatqc-python-scripts/pulls/) that they have run fine...

jsturdy (changes requested):
Looking at the code, the reason that the `preCmd` was not populated at the beginning was because not all tools would be expecting it, e.g., `ultraThreshold`, `ultraLatency`, `fastLatency`
This is to prevent running a configuration for a system that had previously been configured.  Best would be to revert that change and then evaluate whether the operational case needs `confChamber` for any type of tool that `run_scans` will execute

bdorney:
Changes reverted, but shelf functionality still present.

For `ultraLatency.py` and `fastLatency.py` offhand I would expect it would be necessary to change VT1 and trim configuration for a given scan of interest.   On further investigation it seems these two scripts take some parameters but only use a few of them (e.g. they both take `--vt1` argument but they don't ever use it).  While they do use *some* parameters they are passed (e.g. `--vt2` in the case of `ultraLatency.py`).

Additionally for `ultraThreshold.py` you might want be interested in scanning with different mspl values.  But it does not appear to be used.

I think this highlights a level of disorganization we have inserted.  I would propose we move forward in a different way:

`run_scans.py` should set the configuration for each scan using the `--config` flag and `preCmd` list.  Then the scan application which is launched by `run_scans.py` should not do any configuration outside what is necessary for the scan.  For example the following scans appear to do extra configuration:
- `ultraLatency.py` sets `vt2`, `mspl`
- `ultraThreshold.py` sets `vt2`, `latency`
- `fastLatency.py` sets `mspl`

For easier maintenance and organization we should have `run_scans.py` call `confChamber.py` for the link of interest with the desired config and then the scan application should run.

Suggest to make this a new issue and include in a separate PR.

run_scans.py
```diff
@@ -21,7 +21,7 @@ def launchTestsArgs(tool, shelf, slot, link, chamber, vfatmask, scanmin, scanmax
 
   #Build Commands
   setupCmds = []
-  preCmd = None
+  preCmd = None
```
jsturdy: there appears to be some trailing whitespace here (L24)

run_scans.py
```diff
@@ -132,7 +132,8 @@ def launchTestsArgs(tool, shelf, slot, link, chamber, vfatmask, scanmin, scanmax
       pass
     log = file("%s/scanLog.log"%(dirPath),"w")
     if preCmd and config:
-      runCommand(preCmd,log)
+      #runCommand(preCmd,log)
+      runCommand(preCmd)
```
jsturdy: Any reason why you removed the logging (or just commented it out)?

jsturdy (changes requested):

jsturdy:
Right, when all update functionality has been added, we can then start to reorganize and package

bdorney:
Testing updates: http://cmsonline.cern.ch/cms-elog/1005667

Also as a by-product when using the `--debug` the `vfatmask` will now be correctly written in hex.  The reason it was not being written as hex previously seems to be some weird interaction of the `dict.values()` method.  Calling `chamber_vfatmask.values()` returns a `list` where all the entries are formatted as `int`'s in base 10.  Using the `hex()` function and python list comprehension was able to solve this.

run_scans.py
jsturdy: Still there...

run_scans.py
jsturdy: ?

jsturdy (changes requested):

run_scans.py
jsturdy: Removing the logging, is not the correct thing to do here.  If there's a problem, the problem should be addressed.
Logging in this manner has been working for a long time

jsturdy (changes requested):

run_scans.py
bdorney: Searching for `\s` characters in vim.  Doesn't show up:

screen shot 2017-08-31 at 11 19 24

bdorney: It may be that changes weren't pushed to the remote...?

jsturdy: Agree, current file on github appears fixed

run_scans.py
jsturdy: Still needs resolution

jsturdy (changes requested):

run_scans.py
jsturdy: NVM, see update now

jsturdy (approved):

bdorney:
Looks like again `code-quality/landscape` is stuck in `Waiting for status to be reported` for ~9 days now...this seems to be a re-occurring issue.

Issue #122, opened 2017-09-01
Feature Request: Cross-check Configuration in confChamber.py


bdorney:
## Brief summary of issue

Few times while performing latency scans the scan is taken at VT1=100 for all VFATs even though a custom configuration is used by supplying the `--vfatConfig` option to `confChamber.py`.  See for example:

http://cmsonline.cern.ch/cms-elog/1007249

After doing some checks in the scan module in the FW does not seem to be the problem (http://cmsonline.cern.ch/cms-elog/1007411).

Suggest to add a cross-check of what is written to `confChamber.py`

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

## Expected Behavior

When supplying the `--vfatConfig` option `confChamber.py` should compare, after writing, the readback values with the supplied values and print to the user if there are any differences.  Maybe this should be extended to `--chConfig` option however there are ~3072 registers to check so that might increase time "too much".

## Current Behavior

There is no check of what is written in `confChamber.py` when the `--vfatConfig` or `--chConfig` registers are used.

## Context (for feature requests)

Sometimes latency scans have been taken and the `VT1` value that is stored in the outptu `latTree` does not meet expectations, and is 100.  Seems to indicate that the default VT1 value is still stored and not the requested values provided by the `--vfatConfig` option.

## Your Environment

* Version used: 18496a1506f0312d2834062bce2e5708abc6d583
* Shell used: /bin/bash

PR #124, opened 2017-09-01
New Feature: Cross-Check of Registers in confChamber.py (#122)

bdorney:
## Description

Implementation of: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/122

Adds to `confChamber.py` the `--compare` option.  The following behavior occurs:

- Now by default whenever `--filename`, `--chConfig`, and/or `--vfatConfig` are supplied a readback of registers is performed and reports any differences between the values that are stored, after writing, and the values requested
- If `--compare` option is supplied with `--filename`, `--chConfig`, and/or `--vfatConfig` registers are **not** written, instead a readback is performed and any differences between values that are stored and the values requested are reported.

For `--filename` or `--chConfig` the readback is of VFAT channel registers and `trimDAC` and `mask` are checked.

For `--vfatConfig` the readback is of `VThreshold1` and `ContReg3` (e.g. `trimRange`).

### Types of changes

## Motivation and Context

Provides a cross-check of the configuration to ensure the correct register values are written.

## How Has This Been Tested?

Freshly tested here: http://cmsonline.cern.ch/cms-elog/1009166

## Checklist:

### DIFF
--- a/confChamber.py
+++ b/confChamber.py
@@ -26,6 +26,8 @@ parser.add_option("--vt1", type="int", dest="vt1",
                   help="VThreshold1 DAC value for all VFATs", metavar="vt1", default=100)
 parser.add_option("--vt1bump", type="int", dest="vt1bump",
                   help="VThreshold1 DAC bump value for all VFATs", metavar="vt1bump", default=0)
+#parser.add_option("-w","--write",action="store_true", dest="write",
+#                  help="Write the current settings to a file", metavar="write")
 
 (options, args) = parser.parse_args()
 
@@ -71,6 +73,12 @@ if options.filename:
         for event in inF.scurveFitTree :
             writeVFAT(ohboard,options.gtx,int(event.vfatN),"VFATChannels.ChanReg%d"%(int(event.vfatCH)),int(event.trimDAC)+32*int(event.mask))
             writeVFAT(ohboard, options.gtx, int(event.vfatN), "ContReg3", int(event.trimRange),0)
+        
+        if options.verify:
+            chTree = inF.Get("scurveFitTree")
+            dict_readBack = { "trimDAC":"VFATChannels.ChanReg", "mask":"VFATChannels.ChanReg" }
+            readBackCheck(chTree, dict_readBack, ohboard, options.gtx)
+
     except Exception as e:
         print '%s does not seem to exist'%options.filename
         print e
@@ -83,6 +91,11 @@ if options.chConfig:
 
         for event in chTree :
             writeVFAT(ohboard,options.gtx,int(event.vfatN),"VFATChannels.ChanReg%d"%(int(event.vfatCH)),int(event.trimDAC)+32*int(event.mask))
+        
+        if options.verify:
+            dict_readBack = { "trimDAC":"VFATChannels.ChanReg", "mask":"VFATChannels.ChanReg" }
+            readBackCheck(chTree, dict_readBack, ohboard, options.gtx)
+
     except Exception as e:
         print '%s does not seem to exist'%options.filename
         print e
@@ -98,14 +111,14 @@ if options.vfatConfig:
             writeVFAT(ohboard, options.gtx, int(event.vfatN), "VThreshold1", int(event.vt1+options.vt1bump),0)
             writeVFAT(ohboard, options.gtx, int(event.vfatN), "ContReg3", int(event.trimRange),0)
 
+        if options.verify:
+            if options.vt1bump != 0:
+                print "Mismatches between write & readback valus for VThreshold1 should be exactly %i" %(options.vt1bump)
+            dict_readBack = { "vt1":"VThreshold1", "trimRange":"ContReg3" }
+            readBackCheck(vfatTree, dict_readBack, ohboard, options.gtx)
+
     except Exception as e:
         print '%s does not seem to exist'%options.filename
         print e
         
-if options.verify:
-    if options.vt1bump != 0:
-        print "Mismatches between write & readback valus for VThreshold1 should be exactly %i" %(options.vt1bump)
-    dict_readBack = { "vt1":"VThreshold1", "trimRange":"ContReg3" }
-    readBackCheck(vfatTree, dict_readBack, ohboard, options.gtx)
-
 print 'Chamber Configured'
--- a/confChamber.py
+++ b/confChamber.py
@@ -70,18 +70,15 @@ if options.filename:
         chTree = inF.Get("scurveFitTree")
         dict_readBack = { "trimDAC":"VFATChannels.ChanReg", "mask":"VFATChannels.ChanReg" }
 
-        if options.compare:
-            print 'Comparing Currently Stored Channel Registers with %s'%options.filename
-
-            readBackCheck(chTree, dict_readBack, ohboard, options.gtx)
-        else:
+        if not options.compare:
             print 'Configuring Channel Registers based on %s'%options.filename
             
-            for event in inF.scurveFitTree :
+            for event in inF.scurveFitTree:
                 writeVFAT(ohboard,options.gtx,int(event.vfatN),"VFATChannels.ChanReg%d"%(int(event.vfatCH)),int(event.trimDAC)+32*int(event.mask))
                 writeVFAT(ohboard, options.gtx, int(event.vfatN), "ContReg3", int(event.trimRange),0)
-        
-            readBackCheck(chTree, dict_readBack, ohboard, options.gtx)
+ 
+        print 'Comparing Currently Stored Channel Registers with %s'%options.filename
+        readBackCheck(chTree, dict_readBack, ohboard, options.gtx)
 
     except Exception as e:
         print '%s does not seem to exist'%options.filename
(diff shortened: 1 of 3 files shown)


confChamber.py
```diff
@@ -93,6 +97,13 @@
             print 'Set link %d VFAT%d VThreshold1 to %i'%(options.gtx,event.vfatN,event.vt1+options.vt1bump)
             writeVFAT(ohboard, options.gtx, int(event.vfatN), "VThreshold1", int(event.vt1+options.vt1bump),0)
             writeVFAT(ohboard, options.gtx, int(event.vfatN), "ContReg3", int(event.trimRange),0)
+
+        if options.verify:
+            if options.vt1bump > 0:
```
jsturdy: I think that at line 102, `vt1bump` *could* be negative (this "feature" has been used in the past)

qcutilities.py
```diff
@@ -0,0 +1,38 @@
...
+
+import numpy as np
+import root_numpy as rp
+import ROOT as r
+import sys
+
+# rootTree - name of a TTree that has values that should have been written
+# dict_Names - dictionary where key names are the branch names in rootTree and the values are the register names they correspond too
+# device - optohybrid  the vfats belong to that you want to check
+# gtx - link of this optohybrid
```
jsturdy: You can write this information as a [`docstring`](https://www.python.org/dev/peps/pep-0257/) so that it's a property of the function

confChamber.py
```diff
@@ -3,19 +3,23 @@
...
+from qcutilities import readBackCheck 
 
 parser.add_option("--chConfig", type="string", dest="chConfig", default=None,
                   help="Specify file containing channel settings from anaUltraSCurve", metavar="chConfig")
 parser.add_option("--filename", type="string", dest="filename", default=None,
                   help="Specify file containing settings information", metavar="filename")
 parser.add_option("--run", action="store_true", dest="run",
                   help="Set VFATs to run mode", metavar="run")
+parser.add_option("--verify", action="store_true", dest="verify",
+                  help="Performs a readback of registers and compares stored values with written values", metavar="verify")
```
jsturdy: Can the `verify` option be used standalone?  I.e., one doesn't write the config, but just checks that what are in the chips currently conforms to the expected config?
Probably a little bit of work, or maybe as simple as adding a `write`/`set` option which if true, puts the config into the chips..., but this will probably play into the comment about migrating functionality longer term into `cmsgemos/gempython`

qcutilities.py
jsturdy: A la:
```python
def readBackCheck(rootTree, dict_Names, device, gtx):
"""<short one line description of method>

rootTree - name of a TTree that has values that should have been written
dict_Names - dictionary where key names are the branch names in rootTree and the values are the register names they correspond too
device - optohybrid  the vfats belong to that you want to check
gtx - link of this optohybrid
"""
...
```

jsturdy (changes requested):
I think that the `readBackCheck` should eventually migrate to `cmsgemos/gempython`, as `verifyConfig`
But before that it should be made more flexible in terms of where it gets the config to check against (xml, db, `root` tree, csv...)

confChamber.py
bdorney: Yes I can change the`verify` option to be standalone.  No problem here.

I can also work on adding a `write/set` option for this PR as well.

confChamber.py
bdorney: Thanks for this point, I will change this `if` statement to reflect this possibility.

qcutilities.py
bdorney: Ah okay, thanks for the example.  I will do this.

confChamber.py
bdorney: After thinking about this some more this would require a larger overhaul.  What is it that we want `--verify` to do?  The original idea was that if `--chConfig` and/or `--vfatConfig` was supplied it would cross-check against the input file that reg values that are read match reg values that were asked to be written.

So to have `--verify` as standalone doesn't seem to make sense since it requires an input file in each case.  Maybe you could elaborate more what you'd like it to do?

In the meantime I'll try to come up with a `write/set` option.  For simplicity I think the `write/set` option should produce three text files:

- `chConfig.txt`, as is now.
- `vfatConfig.txt`, as is now.
- `paramConfig.txt`.

Where `paramConfig.txt` is a text file that for each VFAT has the following registers:

```
        "ContReg1":    0x00,
        "ContReg2":    0x30,
        "ContReg3":    0x00,
        "Latency":      169,
        "IPreampIn":    144,
        "IPreampFeed":   69,
        "IPreampOut":   129,
        "IShaper":      130,
        "IShaperFeed":   87,
        "IComp":        101,
        "VCal":           0,
        "VThreshold2": 0x00,
        "CalPhase":    0x00
```

The format of `paramConfig.txt` would then match that of `chConfig.txt` and `vfatConfig.txt` (e.g. a tab delimited file).

as defined in `parameters.defaultValues` of `vfat_user_functions.py`

bdorney:
I think to implement a `write/set` option to write out the config to a file, and set the config from a file we should think this over more and make a separate PR.

This would be a "larger" change as it might invalidate the `--chConfig` and `--vfatConfig` options presently.  Additionally there might be some redundancy in how the VFAT DAC parameters are applied by the `chamber_vfatDACSettings` `dict`.

This should be it's own issue/PR.

bdorney:
> I think that the readBackCheck should eventually migrate to `cmsgemos/gempython`, as `verifyConfig` But before that it should be made more flexible in terms of where it gets the config to check against (xml, db, `root` tree, csv...)

I would need some help with that.  Right now the algorithm is relying on the `root_numpy` method `tree2array` to allow an arbitrary input of register names and values to check against.  This seemed to be the most simplistic since most of our ultra scan applications produce `TTree` objects that store data which can be used for configuration.

If this is the long-term goal then I suggest an issue be opened in `cmsgemos` for this.  However in the short-term I propose the solution presented here.

qcutilities.py
```diff
@@ -5,14 +5,20 @@
 import ROOT as r
```
jsturdy: It's probably better to put this/these imports into the function that needs them

jsturdy (changes requested):
It's not clear to me from the code whether the `--verify` option can be used without writing anything to the chips, i.e., as simply a cross check against some configuration and the current chips.
Can you comment on this (not a stopper, just want to know)

confChamber.py
jsturdy: Yeah, my thought was you could tell it what configuration *should* be loaded into the chips, and check that it in fact is.  It would take the same arguments as the normal `conf/write/set` process, and could be passed in along with the `conf/write/set` option to verify that the newly written config matches expectation of what was just written as well.
I think this answers my question in the review, so no need to reply to that, just fix the one change and it should be good to go.

qcutilities.py
bdorney: moved

bdorney:
> It's not clear to me from the code whether the --verify option can be used without writing anything to the chips, i.e., as simply a cross check against some configuration and the current chips.
> Can you comment on this (not a stopper, just want to know)

I think it could be but we would need an additional option or to change the implementation.  Right now `--verify` only does something if `--filename`, `--chConfig`, and/or `--vfatConfig` are supplied.  When these last three options are supplied right now by default writes occur (since it is assumed if you are giving those options you want the settings defined in those files written to the chips).

I see a few ways forward:

1. Change the default behavior such that a readback is *always* performed, but if the `--verify` option is supplied then `write` of VFAT registers that corresponds to the supplied input file {`--filename`,`chConfig`,`vfatConfig`} does **not** occur and only the readback is performed; or
2. Calling `--verify` itself suppresses the `write` and only performs the readback. 
2. Keep the current implementation but add an additional option `--readOnly` which if supplied *with* `--verify` would suppress the writing and perform only the readback.

I would be in favor of option 1 or 2 as it limits option bloat and in my mind is more intuitive to the user.  Option 1 has the disadvantage is that every time we perform double the number of transactions (e.g. write and read for each register).  Option 2 may be the best but it requires the user to first execute a call of `confChamber.py` to `write` the registers and then another call to read them back so this requires more "actions" and may not be desired.  Option 3 may be less intuitive and have more "fiddlie" bits but it would not increase the number of default transactions.

Do you have a preference or suggestion of an alternative?  If not I'll proceed with option 1 above.

jsturdy:
OK, if implementing option 1 above, then semantically `compare` is a better name than `verify`, since the verification will be done for every configuration that is written, but the user may at some future time simply wish to compare a given configuration with the current configuration.
We can forget option 2, for semantics, as well as for the fact that, as your initial request was for a verification that things were written properly, it would break that workflow.
We may at some future time wish to suppress the verify step for every configuration, but it's pretty trivial to go from option 1 to option 3(ish), so I think option 1 is the best near-term solution.

bdorney:
Okay cool, I'll work on Option 1 with `--verify` changed to `--compare`

bdorney:
Changes implemented and description of this PR has been updated, fresh tests have been performed and elog above shows current status of tests

jsturdy (commented):
I think the logic of 
```python
if options.compare:
  readback
else:
  configure
  readback
```
could be simplified with:
```python
if not options.compare:
   configure
readback
```
which would also allow for the trivial transformation to a possible future option of `writeconfig`

bdorney:
Updated the `if-statements` used to match:

```
if not options.compare:
   configure
readback
```

New tests performed [here](http://cmsonline.cern.ch/cms-elog/1009166) and looks like everything came back as expected.

jsturdy (approved):

Issue #126, opened 2017-09-11
Feature Request: Filesystem Interaction Before TFile::Write()


bdorney:
## Brief summary of issue

On cosmicstandtif it seems the automount of the 904 NAS times out after some time.  When this happens if a scan is running the output TFile will be a zombie and have no keys causing a data loss.  This is especially frustrating if the scans take a long time to run due to low trigger rate (e.g. cosmics).

Issue has occurred several times, one has been reported here: http://cmsonline.cern.ch/cms-elog/1008108

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

## Expected Behavior

I see two ways that this could be solved.  

The first way is to perform a filesystem interaction before writing out the output `TFile`.  For example in `ultraLatency.py` placing before [L#269](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/18496a1506f0312d2834062bce2e5708abc6d583/ultraLatency.py#L269):

```
    from gempython.utils.wrappers import runCommand
    runCommand(["stat",filename])
```

The second way could be to:

1. write the file to a temporary directory on the machine, 
2. check the requested output directory exists, and then
3. move the output file to the output directory at the scan end.

I think the first option is simpler and has already been tested with a local change at:

http://cmsonline.cern.ch/cms-elog/1008108
http://cmsonline.cern.ch/cms-elog/1009162

While latency scans are potentially the longest ultra scans we have right now I would suggest adding this to all of the scan modules:

- `fastLatency.py`,
- `ultraLatency.py`,
- `ultraScurve.py`, and
- `ultraThreshold.py`.

## Current Behavior

If the 904 NAS times out during a scan on cosmicstandtif data loss ensues. 

## Context (for feature requests)

Losing a lot of time and data due to low trigger rate (cosmics).

## Your Environment

* Version used: 18496a1506f0312d2834062bce2e5708abc6d583
* Shell used: /bin/bash

mexanick:
Becomes irrelevant after migration to CC7

PR #127, opened 2017-09-11
Feature Request: Addressing #126, Filesystem Interaction Before Output Written

bdorney:
## Description

Addressing: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/126

The following scans:
- `fastLatency.py`,
- `ultraLatency.py`,
- `ultraScurve.py`, and
- `ultraThreshold.py`.

Will now perform a filesystem interaction before writing the output file.  Should prevent data loss if `automount` of a NAS on a DAQ machine times out.

### Types of changes

## Motivation and Context

Data loss has been [reported](http://cmsonline.cern.ch/cms-elog/1008108) due to time out of `automount` of a NAS on a DAQ machine.

## How Has This Been Tested?

http://cmsonline.cern.ch/cms-elog/1008108
http://cmsonline.cern.ch/cms-elog/1009162

## Checklist:


jsturdy:
Any update on whether this solution has alleviated the presumed issue?

bdorney:
scandate=2017.09.13.16.45 which went overnight worked, http://cmsonline.cern.ch/cms-elog/100954

Previous scan where I tried this (2017.09.12.23.30, http://cmsonline.cern.ch/cms-elog/1009433) failed for a different reason (SCA issue).

So based on 2017.09.13.16.45 I think it's "working".  Or I just got lucky...launched another scan this morning.

cbravo135 (commented):
I see you are running stat, but it doesn't seem like you look to see if it actually worked. As is, it doesn't seem like it does anything really. I would think the point of this is to catch if the FS is mounted or not if it isn't switch to write to /tmp or have some other way of recovering.

bdorney:
The original idea that @jsturdy presented was that a filesystem interaction would trigger the `automount` so this would either:

- Work by default, or
- Cause `automount` to trigger ensuring the FS is mounted.

However, I think you raise a good point.  It might be worth checking the returncode of the command and then implementing a switch to:

1. Writing to a tmp directory
2. copying output from the tmp directory to the output data path

I think part of the problem also is that the output `TFile` is opened at the start of the scan so if the connection is lost maybe the `TFile` also becomes corrupted...suggest to move the initialization of the output `TFile` to just before the output `TTree` object is written.

cbravo135:
I am just trying to think of this as a problem in general. We shouldn't be telling remote sites how to setup the FS on their machines but we can check and tell them something is wrong and make use of /tmp if we see a potential problem.

jsturdy:
I suppose this hasn't moved due to other issues, or is there any sort of verification that any of these solutions are helping with the misbehaving setup?
Regarding the `TFile` initialization, it can be done in stages, i.e.,
* First `TFile(filename,"RECREATE")`, then close when operations are finished
* Next call `TFile(filename,"UPDATE")`, do new writes, saves, etc., then close

jsturdy:
Closing as "On Hold" to ensure landscape queues don't trigger on pushes to this branch.  Reopen when ready for review

Issue #128, opened 2017-09-14
Bug Report: ultraLatency.py Does Not Record Correct vth1 Values?


bdorney:
## Brief summary of issue

Stumbled on a potential bug.  It looks like `vth1` is not correctly being written to `latTree` by `ultraLatency.py` 

In the latency scan reported at this [E-Log](http://cmsonline.cern.ch/cms-elog/1009271) with `scandate=2017.09.12.14.24`.

The `--vfatConfig` used is:
```
/data/bigdisk/GEM-Data-Taking/GE11_QC8/configs/z4.0/vfatConfig_GE11-VI-L-CERN-0001.txt
```

Here `vt1bump=0`.  

In this case I ran `vfat_info_uhal.py` after configuring, but before running the scan, to xcheck the configuration (this was after fixing [this bug](https://github.com/cms-gem-daq-project/cmsgemos/pull/141)).  In this case `VThreshold1` register values match what is in `--vfatConfig`.

However when the scan completed the `vth1` values stored in `latTree` are incorrect from VFAT10 onward.
<img width="478" alt="treedraw_vth1_vs_vfatn_scandate2017 09 12 14 24" src="https://user-images.githubusercontent.com/6432141/30415582-f4e61582-9928-11e7-945e-c1bb0bcd78da.png">

Specifically:

| VFAT | vfatConfig | Elog (Hex) | Elog (Dec) | latTree |
| ---- | ---------- | ---------- | ---------- | ------- |
| 0 | 104 | 0x68 | 104 | 104 |
| 1 | 51 | 0x33 | 51 | 51 |
| 2 | 70 | 0x46 | 70 | 70 |
| 3 | 82 | 0x52 | 82 | 82 |
| 4 | 110 | 0x6e | 110 | 110 |
| 5 | 113 | 0x71 | 113 | 113 |
| 6 | 119 | 0x77 | 119 | 119 |
| 7 | 139 | 0x8b | 139 | 139 |
| 8 | 38 | 0x26 | 38 | 38 |
| 9 | 42 | 0x2a | 42 | 42 |
| 10 | 0 | 0x0 | 0 | 54 |
| 11 | 54 | 0x36 | 54 | 58 |
| 12 | 58 | 0x3a | 58 | 69 |
| 13 | 60 | 0x3c | 60 | 74 |
| 14 | 74 | 0x4a | 74 | 111 |
| 15 | 111 | 0x6f | 111 | 56 |
[...]

PR #132, opened 2017-09-26
Bug Fix: Remove vfatmask usage when using readAllVFATs(...)

bdorney:
## Description

Addressing: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/128

Based on the discussion shown in [cmsgemos #146](https://github.com/cms-gem-daq-project/cmsgemos/issues/146) it seems like the best approach is to not use a vfatmask when using `readAllVFATs(...)` which does a broadcast read.  This will ensure that the ordering of the returned data is correct by default.

### Types of changes

## Motivation and Context

Values returned by a broadcast read when a vfatmask is used appear out of order.

## How Has This Been Tested?

Change is trivial, and tests performed in [cmsgemos #146](https://github.com/cms-gem-daq-project/cmsgemos/issues/146) show the desired behavior, e.g. supplying a mask of 0x0 returns the correctly ordered data and does not crash.

## Checklist:

- [X] My change requires a change to the documentation.

### DIFF
--- a/fastLatency.py
+++ b/fastLatency.py
@@ -71,15 +71,15 @@ try:
     writeAllVFATs(ohboard, options.gtx, "ContReg2",    ((options.MSPL-1)<<4))
     # writeAllVFATs(ohboard, options.gtx, "VThreshold1", options.vt1)
 
-    vals  = readAllVFATs(ohboard, options.gtx, "VThreshold1", mask)
+    vals  = readAllVFATs(ohboard, options.gtx, "VThreshold1", 0x0)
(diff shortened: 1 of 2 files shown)


jsturdy (approved):
Looks OK, but I can't say there are nested calls where you've passed in `mask` and it gets forwarded...
There are also a couple of `write` transactions which have the same `mask` (this won't have any of the issues that you encountered with the `read` transactions, so it isn't really a problem, more of a note in case this hadn't been considered)

PR #136, opened 2017-10-07
Update requirements.txt

jsturdy:
## Description
fix bug in `requirements.txt` where `root-numpy` was used instead of `root_numpy` failing `pip` installation step


bdorney (approved):
Thanks for catching this typo, my bad.

PR #141, opened 2017-10-20
New Feature: ChipID added to TTree, readBackCheck now checks ChipID & vt1bump 

bdorney:
## Description

Addresses: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/138

Requires: https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/51

Additionally when `confChamber.py` calls `readBackCheck(...)` it will now check the ChipID of the VFATs being configured against what is supplied in the input config files.

Note that as a result new configuration files (e.g. `vfatConfig.txt` and `chConfig.txt`) would need to be generated which have a `ChipID` column (shown in integer representation).

### Types of changes

## Motivation and Context

As we move forward eventually we will be configuring with the information that will be stored in the database.  We should be able to cross-check that the info stored for a given VFAT in the database matches the VFAT that is being configured on the hardware.  The best way to do this is to compare the supplied `vfatID` with the `ChipID1` register on the VFAT since they should be identical.

Additionally in our calibration data there is no way to know which VFATs were used for a given run.  Now all produced TTree's by our calibration scans will include the `vfatID` branch which stores the value of the `ChipID1` register. 

## How Has This Been Tested?

Took new measurements, analyzed them, and configured from the produced config files as shown:

http://cmsonline.cern.ch/cms-elog/1016357
http://cmsonline.cern.ch/cms-elog/1016594

## Checklist:

### DIFF
--- a/confChamber.py
+++ b/confChamber.py
@@ -68,7 +68,7 @@ if options.filename:
     try:
         inF = r.TFile(options.filename)
         chTree = inF.Get("scurveFitTree")
-        dict_readBack = { "trimDAC":"VFATChannels.ChanReg", "mask":"VFATChannels.ChanReg", "vfatID:ChipID1" }
+        dict_readBack = { "trimDAC":"VFATChannels.ChanReg", "mask":"VFATChannels.ChanReg", "vfatID":"ChipID1" }
 
         if not options.compare:
             print 'Configuring Channel Registers based on %s'%options.filename
@@ -88,7 +88,7 @@ if options.chConfig:
     try:
         chTree = r.TTree('chTree','Tree holding Channel Configuration Parameters')
         chTree.ReadFile(options.chConfig)
-        dict_readBack = { "trimDAC":"VFATChannels.ChanReg", "mask":"VFATChannels.ChanReg", "vfatID:ChipID1" }
+        dict_readBack = { "trimDAC":"VFATChannels.ChanReg", "mask":"VFATChannels.ChanReg", "vfatID":"ChipID1" }
 
         if not options.compare:
             print 'Configuring Channel Registers based on %s'%options.chConfig
@@ -107,7 +107,7 @@ if options.vfatConfig:
     try:
         vfatTree = r.TTree('vfatTree','Tree holding VFAT Configuration Parameters')
         vfatTree.ReadFile(options.vfatConfig)
-        dict_readBack = { "vt1":"VThreshold1", "trimRange":"ContReg3", "vfatID:ChipID1" }
+        dict_readBack = { "vt1":"VThreshold1", "trimRange":"ContReg3", "vfatID":"ChipID1" }
 
         if not options.compare:
             print 'Configuring VFAT Registers based on %s'%options.vfatConfig
--- a/confChamber.py
+++ b/confChamber.py
@@ -68,7 +68,7 @@ if options.filename:
     try:
         inF = r.TFile(options.filename)
         chTree = inF.Get("scurveFitTree")
-        dict_readBack = { "trimDAC":"VFATChannels.ChanReg", "mask":"VFATChannels.ChanReg" }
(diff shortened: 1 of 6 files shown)


jsturdy:
Any reason to not use the full chipID in both writing and checking?
Don't know if we are certain of uniqueness of only one set of the 8 bits in any given chamber

bdorney:
> Any reason to not use the full chipID in both writing and checking? Don't know if we are certain of uniqueness of only one set of the 8 bits in any given chamber.

I did not know there was additional information to use beyond what was returned by `getChipID(...)` or `getAllChipIDs(...)`.  How is this additional information accessed?  

Moreover how does it compare with what is returned by `vfat_info_uhal.py`?  Right now if the value stored by the `vfatID` is converted to hex it will match with what is returned by a call of `vfat_info_uhal.py`.  As can be seen by the elog post when I purposefully put in a wrong vfatID to test `readBackCheck(...)`:

> VFAT22: ChipID1 mismatch, expected = 0xf9e2, readback = 0xf9e4

jsturdy:
OK, so it's *not* `ChipID1` it' the full `chipID`?
I saw some references in the code to the `ChipID1` register, but the full VFAT `chipID` is encoded in 3 (2 used) 8-bit registers, so if you're only using one of them, you'll be opening yourself up to mismatches and overlaps

bdorney:
> OK, so it's not ChipID1 it' the full chipID? I saw some references in the code to the ChipID1 register, but the full VFAT chipID is encoded in 3 (2 used) 8-bit registers, so if you're only using one of them, you'll be opening yourself up to mismatches and overlaps

I think there's some confusion/ignorance on my part then. What is [getChipID](https://github.com/cms-gem-daq-project/cmsgemos/blob/9e731bb89ff9b7308b05380a803a1bf166642feb/gempython/tools/vfat_user_functions_uhal.py#L272) returning here:

```
elif (((thechipid>>16)&emptyMask) != ((0x050<<4)+(chip))):
        thechipid  = (thechipid & ebmask)<<8
        thechipid |= (readVFAT(device, gtx, chip, "ChipID0") & ebmask)
```

Is this the full chipID?  If not, why not?  How does it differ from what is being returned by [getAllChipIDs](https://github.com/cms-gem-daq-project/cmsgemos/blob/9e731bb89ff9b7308b05380a803a1bf166642feb/gempython/tools/vfat_user_functions_uhal.py#L288).

Right now the `TTree` produced in the scan applications are storing the value returned by `getChipID(...)` and the comparison in `readBackCheck(...)` is done by comparing this stored value with what is returned by `getAllChipIDs(...)` 

If this is insufficient please advise as to what register names should be used.  Additionally I would suggest to amend these functions to return complete information if they are not presently doing so.

confChamber.py
```diff
@@ -107,7 +107,7 @@
     try:
         vfatTree = r.TTree('vfatTree','Tree holding VFAT Configuration Parameters')
         vfatTree.ReadFile(options.vfatConfig)
-        dict_readBack = { "vt1":"VThreshold1", "trimRange":"ContReg3" }
+        dict_readBack = { "vt1":"VThreshold1", "trimRange":"ContReg3", "vfatID":"ChipID1" }
```
jsturdy: I guess this address isn't used to read then (also l71, l91)?
This was where I was unclear whether you were getting just the 8 bits corresponding to `ChipID1` or the full `chipID`, as returned by `getChipID`

jsturdy (dismissed):

confChamber.py
bdorney: if it would be more clear I can change the string here from `ChipID1` to just `ChipID`

jsturdy: Probably better

bdorney: Done

jsturdy (approved):

PR #144, opened 2017-10-23
New Feature: Common TTree Format of Calibration Scans

bdorney:
## Description

Eventually we'd like to store the results of the calibration scans in the DB.  I envision that this will lead to a table were numeric data will be stored.  To simplify the work of the DB team I propose a common TTree structure for GEM scans be used for:

- `fastLatency.py`
- `ultraLatency.py`
- `ultraScurve.py`
- `ultraThreshold.py`

The new TTree is called `gemTree` and is defined as a data member of the `gemTreeStructure` class of `treeStructure.py`.  Since the above scans have a very similar structure the proposed TTree is simply just all the TBranches of the above scan types.  

In the case where a particular branch does not apply to a given scan it is initialized with a dummy value of either 0 or -1 (following what is used currently).

There are two breaking changes:
- The `lat` branch of that was previously produced by `ultraLatency.py` has been renamed to `latency` to match what was being previously produced by `ultraScurve.py`.
- The `vthr` branch that was previously produced by `ultraScurve.py` has been renamed to `vth1` to match what was being previously produced by `ultraLatency.py` and `ultraThreshold.py`

### Types of changes

## Motivation and Context

In the future I would like the results of our calibration scans stored in the DB as a table of numbers (and NOT a TFile) such that the WBM can easily make the calibration plots rather than the DOC/DPG contact.

## How Has This Been Tested?

I took new scans and analyzed them, updating for the branch name changes above.  See E-Log and screen shots shown here:

http://cmsonline.cern.ch/cms-elog/1017334

## Checklist:

### DIFF
--- a/fastLatency.py
+++ b/fastLatency.py
@@ -36,13 +36,11 @@ printSystemTTCInfo(amcBoard, options.debug)
 from ROOT import TFile,TTree
 filename = options.filename
 myF = TFile(filename,'recreate')
-myT = TTree('latencyTree','Tree Holding CMS GEM Latency Data')
 
 # Setup the output TTree
 from treeStructure import gemTreeStructure
-gemData = gemTreeStructure('latencyTree','Tree Holding CMS GEM Latency Data')
-gemData.link[0] = options.gtx
-gemData.Nev[0] = options.nevts
+gemData = gemTreeStructure('latencyTree','Tree Holding CMS GEM Latency Data',scanmode.LATENCY)
+gemData.setDefaults(options)
 gemData.vth2[0] = options.vt2
 
 import time
@@ -95,13 +93,13 @@ try:
                 gemData.vth2[0]  = vt2vals[vfat]
                 gemData.vth[0]   = vthvals[vfat]
                 writeRegister(ohboard,"%s.VFAT%d_LAT_BX.RESET"%(baseNode,vfat),0x1)
-                gemData.gemTree.Fill()
+                gemData.fill()
                 pass
             if (seenTriggers%100 == 0):
                 print "Saw %d triggers"%(seenTriggers)
                 pass
             pass
-        myT.AutoSave("SaveSelf")
+        gemData.autoSave()
         sys.stdout.flush()
         if seenTriggers > options.nevts:
             print "Saw %d triggers, exiting"%(seenTriggers)
@@ -109,10 +107,10 @@ try:
             break
 
 except Exception as e:
-    myT.AutoSave("SaveSelf")
+    gemData.autoSave()
     print "An exception occurred", e
     sys.stdout.flush()
 finally:
     myF.cd()
-    gemData.gemTree.Fill()
+    gemData.write()
     myF.Close()
--- a/fastLatency.py
+++ b/fastLatency.py
@@ -33,26 +33,15 @@ filename = options.filename
 myF = TFile(filename,'recreate')
 myT = TTree('latencyTree','Tree Holding CMS GEM Latency Data')
 
-Dly = array( 'i', [ -1 ] )
-myT.Branch( 'Dly', Dly, 'Dly/I' )
-vfatN = array( 'i', [ -1 ] )
-myT.Branch( 'vfatN', vfatN, 'vfatN/I' )
-vth = array( 'i', [ 0 ] )
-myT.Branch( 'vth', vth, 'vth/I' )
-vth1 = array( 'i', [ 0 ] )
-myT.Branch( 'vth1', vth1, 'vth1/I' )
-vth2 = array( 'i', [ 0 ] )
-myT.Branch( 'vth2', vth2, 'vth2/I' )
-mspl = array( 'i', [ -1 ] )
-myT.Branch( 'mspl', mspl, 'mspl/I' )
-link = array( 'i', [ -1 ] )
-myT.Branch( 'link', link, 'link/I' )
-link[0] = options.gtx
-utime = array( 'i', [ 0 ] )
-myT.Branch( 'utime', utime, 'utime/I' )
+# Setup the output TTree
+from treeStructure import gemTreeStructure
+gemData = gemTreeStructure('latencyTree','Tree Holding CMS GEM Latency Data')
+gemData.link[0] = options.gtx
+gemData.Nev[0] = options.nevts
+gemData.vth2[0] = options.vt2
 
 import time
-utime[0] = int(time.time())
+gemData.utime[0] = int(time.time())
 
 ohboard      = getOHObject(options.slot,options.gtx,options.shelf,options.debug)
 seenTriggers = 0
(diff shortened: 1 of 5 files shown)


shaikhhasib:
When you take the calibration scan, the final output is one single root file  for example (LatencyData_Trimmed.root)?? Inside this root file you have three different sets of plot (As I see in Elog) ??  If this is the case then we just need one blob filed where with will store this root file data as blob, when the user request for specific calibration run on OMS page, Resthub API retrieve this blob and run the the pyroot script to open this root file and display on OMS page.  So the question is how each calibration scan is different from each other by run number??

bdorney:
> When you take the calibration scan, the final output is one single root file for example (LatencyData_Trimmed.root)?? 

The output of a calibration scan is a `TFile` which contains a `TTree`, right now the three tools `ultra*.py` and `fastLatency.py` each produce a different `TTree` however many of the branches are common and overlap.

In this proposed change the calibration scans will all produce the *same* TTree object such that a single type of table in the DB can store the different calibration data.  Here we expect each row in the table to be an event in the `TTree` and the columns are the branch names.  So we will not store the `TFile` but the numeric data in the `TTree`.  We can provide a text file that is generated from the `TTree`.

> Inside this root file you have three different sets of plot (As I see in Elog) ?? 

No there will only be a `TTree` the plots should be made by the WBM adapting the current analysis tools available in the `gem-plotting-tools` repository.

> So the question is how each calibration scan is different from each other by run number??

The `mode` branch of the `TTree` provides a unique identifier for the type of run that is taken.  We can also provide a general flag.  The output data will either represent a:

- Latency scan,
- Scurve scan, or
- Threshold scan.

The TBranches will all be the same however some may have zero or dummy values.

fastLatency.py
```diff
@@ -120,5 +109,5 @@
     sys.stdout.flush()
 finally:
     myF.cd()
-    myT.Write()
+    gemData.gemTree.Fill()
```
jsturdy: Should be a `Write` here

shaikhhasib:
If you point me to this TFile, It will help me to understand the structure better

ultraScurve.py
```diff
@@ -140,51 +98,52 @@
             trimVal = (0x3f & readVFAT(ohboard,options.gtx,vfat,"VFATChannels.ChanReg%d"%(scCH)))
             writeVFAT(ohboard,options.gtx,vfat,"VFATChannels.ChanReg%d"%(scCH),trimVal)
 
+    gemData.mode[0] = scanmode.SCURVE
```
jsturdy: Consider having the `gemData` constructor to take the `scanmode.<TYPE>` as a parameter

ultraScurve.py
```diff
@@ -140,51 +98,52 @@
...
     for scCH in range(CHAN_MIN,CHAN_MAX):
-        vfatCH[0] = scCH
+        gemData.vfatCH[0] = scCH
         print "Channel #"+str(scCH)
         for vfat in range(0,24):
             if (mask >> vfat) & 0x1: continue
             trimVal = (0x3f & readVFAT(ohboard,options.gtx,vfat,"VFATChannels.ChanReg%d"%(scCH)))
             writeVFAT(ohboard,options.gtx,vfat,"VFATChannels.ChanReg%d"%(scCH),trimVal+64)
-        configureScanModule(ohboard, options.gtx, scanmode.SCURVE, mask, channel = scCH,
+        configureScanModule(ohboard, options.gtx, gemData.mode[0], mask, channel = scCH,
```
jsturdy: Might consider (with the functions mentioned in the summary) leaving this as `scanmode.SCURVE` and changing the `gemData` constructor to take the `scanmode.<TYPE>` as a parameter (instead of line 101)
Leaves this configure function call clear, but still puts the desired information in the output tree

jsturdy (commented):
If you're putting a full class to handle the tree data, maybe it would make sense to add a few functions to the class to handle the filling/writing, rather than assigning the values through the member variable access?
Not sure if it will improve readability, but naively, it seems like it wouldn't be a bad way to go, considering the movement to a class object to store the info anyway

ultraScurve.py
bdorney: Done

ultraScurve.py
bdorney: Done

fastLatency.py
bdorney: Fixed

bdorney:
@MuhammadHasib it doesn't exist yet...this PR has not been approved.

bdorney:
> If you're putting a full class to handle the tree data, maybe it would make sense to add a few functions to the class to handle the filling/writing, rather than assigning the values through the member variable access?

I have added a few get/set and other utility methods.

ultraLatency.py
```diff
@@ -228,15 +225,16 @@
             sys.stdout.flush()
             pass
         for VC in range(LATENCY_MAX-LATENCY_MIN+1):
-            gemData.latency[0]   = int((dataNow[VC] & 0xff000000) >> 24)
-            gemData.Nhits[0] = int(dataNow[VC] & 0xffffff)
+            latency   = int((dataNow[VC] & 0xff000000) >> 24)
+            Nhits = int(dataNow[VC] & 0xffffff)
+            gemData.setScanResults(latency, Nhits)
```
jsturdy: OK, more what I was trying to say was, the `fill`  you have at line 234 could be something akin to:
`fill(vt1=vt1val,vthr=vthrval,lat=latval)`
and in the class you'd define it:
```python
def fill(**kwargs):
    if "vt1" in kwargs:
      self.vt1[0] = kwargs["vt1"]
...
```
This would allow the call to encapsulate all the data, the `dataTree` class would keep everything organized, and it would simplify the `ultraX` code

ultraLatency.py
```diff
@@ -71,9 +71,8 @@
...
 # Setup the output TTree
 from treeStructure import gemTreeStructure
-gemData = gemTreeStructure('latTree','Tree Holding CMS GEM Latency Data')
-gemData.link[0] = options.gtx
-gemData.Nev[0] = options.nevts
+gemData = gemTreeStructure('latTree','Tree Holding CMS GEM Latency Data',scanmode.LATENCY)
+gemData.setDefaults(options)
 
 import subprocess,datetime,time
 gemData.utime[0] = int(time.time())
```
jsturdy: Combine this in the `setDefaults` call?

ultraScurve.py
```diff
@@ -51,13 +51,11 @@
 
 # Setup the output TTree
 from treeStructure import gemTreeStructure
-gemData = gemTreeStructure('scurveTree','Tree Holding CMS GEM SCurve Data')
+gemData = gemTreeStructure('scurveTree','Tree Holding CMS GEM SCurve Data',scanmode.SCURVE)
+gemData.setDefaults(options)
 gemData.calPhase[0] = options.CalPhase
 gemData.l1aTime[0] = options.L1Atime
 gemData.latency[0] = options.latency
```
jsturdy: Probably want to pull these out of the chip a la ultraLatency.py#115 (general comment for chip parameters that are written to the tree)

jsturdy (changes requested):
Few more clarifications/suggestions for improvement

ultraLatency.py
bdorney: Done

ultraLatency.py
bdorney: Implemented, yes this was a better approach.  Thanks for suggestion.

ultraScurve.py
bdorney: I have implemented your suggestion.  Now our scan scripts will read relevant parameters from the chip to store them.

jsturdy (approved):
Looks good!

Issue #120, opened 2017-08-30
Excessive unnecessary subprocess.call usage


mexanick:
## Brief summary of issue

All the unit scripts are called through system calls instead as functions. 
### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

The unit script should look like (~indentation is not preserved by markdown~, now with correct markdown syntax):

```python
def my_function(my_params):
    print "Hello World"
    pass

def main()
    print "Fetch params"
    # fetch params here
    my_function(my_params)

if __name__ == "__main__":
    main()
```

then it can be called from command line as before, but if one needs to call the function `my_function` from another module, it can be done via simple `import`.

## Current Behavior

Instead every time when such a unit script has to be called, the 
`subprocess.call("python my_script.py my parameters")` 
is called.

bdorney:
If I understand correctly what you are proposing is that `anaUltra*.py` go from scripts to functions with a main defined? e.g. for `ultraScurve.py` you would propose to have:

```
def ultraScurve( args ):
    code
    code
    code

    return

if __name__ == "__main__":
    get command line arguments
    ultraScurve( those command line arguments )
```

Correct?

I suppose you propose the same thing for `gem-plotting-tools`?  If so I would suggest to open a similar issue there as well.

Could you highlight or elaborate on the disadvantage of the current implementation?  Specifically:

> Instead every time when such a unit script has to be called, the
subprocess.call("python my_script.py my parameters")
is called.

What does this mean and why is it undesired?

mexanick:
Yes, you're correct about what I want. 
Now, what means "subprocess call involved":

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/1b1944142bd102937c3c71db03023fc7057aecaa/confAllChambers.py#L15-L35

where 

https://github.com/cms-gem-daq-project/cmsgemos/blob/b85f5bb29772760ec22ea772cc02e63dd4623699/gempython/utils/wrappers.py#L2-L22

and this all is happening instead of 
```
from my_module import my_function
my_function(my_args)
```
Considering why the subprocess calls are undesired:
1. It is unnecessary
2. If you will ever want to return some data other then status code, you will have difficulties (it is possible, but way more complicated w.r.t. regular python functions)
3. There's an overhead to invoke the system call. Not that much, but why should we waste resources?
[...]

PR #170, opened 2018-06-01
New Feature [WIP]: Better Trimming for VFAT3

bdorney:
### Previous Trim Algorithm
The previous trim algorithm for vfat3 makes a false assumption.  The strategy was:

1. Take an initial scurve with all channel registers set to zero, 
2. For a given `vfat3` determine the distribution of `scurvePOI = scurve_mu - ztrim * scurve_width`,
3. For the `i^th` scurve determine the value `trim = scruvePOI - <scurvePOI>`,
4. Set the `trim` value to the trim amplitude register of the `i^th` channel, the sign of `trim` determines the trim polarity register,
5. Take a final s-curve to evaluate the trim procedure.

### Flawed Assumption of Previous Trim Algorithm
While this procedure has success in reducing the spread of the `scurvePOI` distribution it makes a flawed assumption.  Namely that the value `trim = scruvePOI - <scurvePOI>` which is in terms of `CFG_CAL_DAC` units (e.g. the step size of this register) corresponds to step size of the `ARM_TRIM_AMPLITUDE`.

The step size of the `CFG_CAL_DAC` register is in `fC` and is provided by an external calibration.  The step size of the `ARM_TRIM_AMPLITUDE` is reported in the `vfat3` manual as having an LSB of 0.5 mV.  So assumptions 4 is not correct.  This also can explain the slight under/overshooting in trimming that was been previously reported when comparing the initial s-curve to the post trim s-curve.

### Correction to Trim Algorithm
The correction that is implemented is to require:

1. The trimming require the user provides an input calibration file (e.g. addition of the `--calFile` argument),
2. The `scurvePOI` values are converted into charge units (fC) by the `ScanDataFitter` class using hte provided calibration file,
3. The `vfat3` manual states that the calibration unit of `vfat3` uses a `100 fF` capacitor to perform the charge injection; therefore the difference `scruvePOI - <scurvePOI>` in fC can be converted to volts via `Q = CV` -> `V = Q / C`
4. Once `scruvePOI - <scurvePOI>` is converted into voltage it can be converted into DAC units of the `ARM_TRIM_AMPLITUDE` register by converting into `mV` and then dividing by the LSB (`0.5 mV`) of the `ARM_TRIM_AMPLITUDE` register.

After arriving at the end of step 4 this will then provide a more accurate trim procedure that should resolve this over-under shoot that has been exhibited and force the distribution of `scruvePOI` to even  narrower post trimming.

### Types of changes

## Motivation and Context

See above discussion.

## How Has This Been Tested?

Testing is in progress.

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ preprpm: default
 	$(MakeDir) $(ScriptDir)
 	@cp -rf run_scans.py   $(ScriptDir)
 	@cp -rf trimChamber.py $(ScriptDir)
+	@cp -rf trimChamberV3.py $(ScriptDir)
 	@cp -rf fastLatency.py $(ScriptDir)
 	@cp -rf ultra*.py      $(ScriptDir)
 	@cp -rf conf*.py       $(ScriptDir)
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ default:
 	$(MakeDir) $(PackageDir)
 	@cp -rf qcoptions.py $(PackageDir)
 	@cp -rf qcutilities.py $(PackageDir)
+	@cp -rf treeStructure.py $(PackageDir)
 	@echo "__path__ = __import__('pkgutil').extend_path(__path__, __name__)" > pkg/$(Namespace)/__init__.py
 	@cp -rf __init__.py $(PackageDir)
 
--- a/confChamber.py
+++ b/confChamber.py
@@ -28,6 +28,8 @@ parser.add_option("--vt2", type="int", dest="vt2",
                   help="VThreshold2 DAC value for all VFATs (v2b electronics only)", metavar="vt2", default=0)
 parser.add_option("--vt1bump", type="int", dest="vt1bump",
                   help="VThreshold1 DAC bump value for all VFATs", metavar="vt1bump", default=0)
+parser.add_option("--zeroChan", action="store_true", dest="zeroChan",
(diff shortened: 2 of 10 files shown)


bdorney:
A similar issue may also happen in `vfat2` but I have not investigated it thoroughly.

bdorney:
Changes in some case now also address: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/120

Needed working latency, threshold, and scurve scans to be able to configure for a trim run so modifications were necessary.  Additionally functionality for setting channel registers in a timely fashion was needed so updates to `confChamber.py` were also performed.

During the packaging re-base it seems also that some mish-mash of `gemData` and `treeStructure.py` were tossed in that were not part of the original branch.  I have resolved the conflicts this has created.

So far updates work as expected:

latency (internal trigger)
latencysummary

threshold per channel
threshsummary

scurve:
summary

VFATs 5 & 12 were masked for all tests, the latency scan was taken with non-optimized `CFG_THR_ARM_DAC` register value just to xcheck the latency to use for trimming and scurves.

The optimizations to `trimChamberV3.py` still need more work as the changes made did not give expected results.

Details can be found: http://cmsonline.cern.ch/cms-elog/1046787

Work in progress continues.

bdorney:
Testing `trimChamberV3.py` still ongoing.  PR is not stale.

PR #187, opened 2018-06-26
Bug Fix [V3]: resolution of post git rebase errors in python tools

bdorney:
## Description

Post `git rebase` seems to have introduced several syntax issues and `ValueError`s dues to automatic merging.  This specifically relates to an instance of `gemTreeStructure` being included without being defined.

As a result switched `ultra*.py` v3 scans to use `gemTreeStructure` and added several branches required to enable v3 functionality (e.g. the `calSF` branch).

Finally the `git rebase` seems to have deleted the sbit rate scan applications.  This is corrected.

Changes in some case now also address: #120.

This is related to #170 but will cover fixes to:

- `confChamber.py`
- `ultra*.py`

### Types of changes

## Motivation and Context

Python scripts were unusable.

@mexanick needs a starting point for the strip-cross-talk script.

## How Has This Been Tested?

Extensive testing, for example see: http://cmsonline.cern.ch/cms-elog/1046787

## Checklist:

- [X] My change requires a change to the documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ preprpm: default
 	$(MakeDir) $(ScriptDir)
 	@cp -rf run_scans.py   $(ScriptDir)
 	@cp -rf trimChamber.py $(ScriptDir)
+	@cp -rf trimChamberV3.py $(ScriptDir)
 	@cp -rf fastLatency.py $(ScriptDir)
 	@cp -rf ultra*.py      $(ScriptDir)
 	@cp -rf conf*.py       $(ScriptDir)
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ default:
 	$(MakeDir) $(PackageDir)
 	@cp -rf qcoptions.py $(PackageDir)
 	@cp -rf qcutilities.py $(PackageDir)
+	@cp -rf treeStructure.py $(PackageDir)
 	@echo "__path__ = __import__('pkgutil').extend_path(__path__, __name__)" > pkg/$(Namespace)/__init__.py
 	@cp -rf __init__.py $(PackageDir)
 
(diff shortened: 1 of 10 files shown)


mexanick:
there's still a problem with Travis build. I think it originates from
```bash
SyntaxError: ('invalid syntax', ('/usr/lib/python2.6/site-packages/gempython/scripts/trimChamberV3.py', 194, 52, '    dict_scurvePOIPerChan = { vfat:np.zeros(128) for vfat in range(0,24) }\n'))
```
If the script works in another environment, I would suggest that the problems is in python/numpy version mismatch, e.g. some syntax changed in newer `numpy` version and thus the code doesn't work.
@bdorney could you please confirm that the script works under your env and let us know which `python` version and which `numpy` version you have?

bdorney:
I think the issue is that `travis` is using an older version of python (`v2.6` it seems).  While there's no issue when trying in `python2.7`:

```
(py2.7) % pip list | grep "numpy" --color
numpy                            1.14.5    
root-numpy                       4.7.3.dev0
(py2.7) % python --version
Python 2.7.5
```

For example:

```
(py2.7) [dorney@gem904qc8daq]~/scratch0/CMS_GEM/CMS_GEM_DAQ% python
Python 2.7.5 (default, Apr 11 2018, 07:36:10) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-28)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> dict_scurvePOIPerChan = { vfat:np.zeros(128) for vfat in range(0,24) }
>>> 
```

Should the correct syntax be:

`dict_scurvePOIPerChan = dict( vfat:np.zeros(128) for vfat in range(0,24) )`

More to the point?  Should we be trying to be backwards compatible with `python2.6`?

mexanick (approved):
The failure of CI build is caused by incorrect `slc6` docker config, which is irrelevant in scope of this PR, so we approve it and will resolve CI build under slc6 later

cbravo135 (approved):

PR #189, opened 2018-07-06
New Functionality [V3]: Access by AMC Name

bdorney:
## Description

Previously `v3` python scripts kept same syntax of `--shelf`, `--slot`, `--gtx` as the `v2b` based `uhal` scripts to preserve functionality for the user.

However this had required a hard coded dictionary `cardLocation` in `amc_user_functions_xhal.py` which declared a dictionary in which they key was a tuple of (`shelf`,`slot`).  This obviously was bad design.

Since the `rpc_connection` just takes the `dns` name of an AMC on the daq machine we have dropped support for the `--shelf` and `--slot` arguments.  They have been removed completely in favor of the `--cardName` (short `-c`) argument.  Explicitly wherever `--shelf=X --slot=Y` was used now use just a single argument.

Requires: https://github.com/cms-gem-daq-project/cmsgemos/pull/184

### Types of changes

## Motivation and Context

Frees us from hard coded tracking of slot/shelf locations.

## How Has This Been Tested?

http://cmsonline.cern.ch/cms-elog/1051188

## Checklist:


cbravo135 (approved):

qcoptions.py
```diff
@@ -1,5 +1,12 @@
-from gempython.utils.standardopts import parser
+from optparse import OptionParser
```
jsturdy, line 1: Is everything that uses `qcoptions` happy with this change?

jsturdy (commented):
I suppose the `HwVFAT` and friends will have a failure of a known type if an incorrect `cardName` is specified?
Otherwise, I'd add a check that the hardware endpoint is reachable (`ping` or something simple like that)

bdorney:
> I suppose the HwVFAT and friends will have a failure of a known type if an incorrect cardName is specified? 
Otherwise, I'd add a check that the hardware endpoint is reachable (ping or something simple like that)

This would be more appropriate in `HwAMC::init(...)` who is actually opening the connection.  I can add it there.

jsturdy:
> we have dropped support for the --shelf and --slot arguments. They have been removed completely in favor of the --cardName (short -c) argument

Also, just thinking forward, I wouldn't *remove* the `shelf` and `slot` support, as these should also resolve in most setups (though they can be disconnected from the current tools if necessary), and will be programmatic

qcoptions.py
bdorney: The only one who is "not" is `ultraLatency.py` since communication with `amc13` requires knowledge of the shelf parameter.  This has been added as a dedicated argument for `ultraLatency.py`.

So everything is fine.

bdorney:
> Also, just thinking forward, I wouldn't remove the shelf and slot support, as these should also resolve in most setups (though they can be disconnected from the current tools if necessary), and will be programmatic

How would you propose to use them then for RPC based functionality.  Here `rpc_connect()` does not use the `shelf` or `slot` value but a network alias/ip-address.

jsturdy:
> How would you propose to use them then for RPC based functionality. Here rpc_connect() does not use the shelf or slot value but a network alias/ip-address.

* Create the hostname: `amc-c02-s04.utca` or `amc-c02-s04` for a `localhost` setup
* For P5, the hostnames are also programmatic, `amc-s2e01-23-03`, though slightly more obfuscated (one needs to know the rack number and the physical location of the uTCA shelf)

Not saying you should implement this lookup now, just that it should be done in the future, and leaving the options in, is in that view.

bdorney:
> Not saying you should implement this lookup now, just that it should be done in the future, and leaving the options in, is in that view.

I think we should re-introduce these options once the lookup is implemented.  Right now the goal is try to converge on a suite users can play with; having dead options may be confusing.

jsturdy:
OK, well since the change was already made...

In future, something like this could/(should?) be implemented:
* Connection related options grouped, and can then be passed to the constructor/called script
* Check first on `cardName`, if not present, parse "standard" options into a hostname

Issue #190, opened 2018-07-06
Feature request: confChamber.py and confAllChambers.py write latency


AndrewLevin:
We would like to be able to set the latency register using the confChamber.py and confAllChambers.py scripts, and we cannot now.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Current Behavior
Currently, the options available to confChamber.py and confAllChambers.py do not include an option like --latency to set the latency register.

```
[gemuser@gem904qc8daq gemdaq]$ confAllChambers.py -h                              
Usage: confAllChambers.py [options]

Options:
  -h, --help            show this help message and exit
  -s slot, --slot=slot  slot in uTCA crate
  -g gtx, --gtx=gtx     GTX on the AMC
  --shelf=shelf         uTCA shelf to access
  -d, --debug           print extra debugging information
  -r, --reset           reset link error counters
  --errors=errorRate    calculate link error rates for N seconds
  --testbeam            fixed IP address for testbeam
  --mspl=MSPL           Specify MSPL. Must be in the range 1-8 (default is 4)
  --nevts=nevts         Number of events to count at each scan point
  --scanmin=scanmin     Minimum value of scan parameter
  --scanmax=scanmax     Maximum value of scan parameter
  --vfatmask=vfatmask   VFATs to be masked in scan & analysis applications
                        (e.g. 0xFFFFF masks all VFATs)
  --ztrim=ztrim         Specify the p value of the trim
  --config              Set Configuration from simple txt files
  --run                 Set VFATs to run mode
  --series              Run tests in series (default is false)
  --vt1=vt1             VThreshold1 DAC value for all VFATs
  --vt1bump=vt1bump     VThreshold1 DAC bump value for all VFATs
[gemuser@gem904qc8daq gemdaq]$ confChamber.py -h           
Usage: confChamber.py [options]

Options:
  -h, --help            show this help message and exit
  -s slot, --slot=slot  slot in uTCA crate
  -g gtx, --gtx=gtx     GTX on the AMC
  --shelf=shelf         uTCA shelf to access
  -d, --debug           print extra debugging information
  -r, --reset           reset link error counters
  --errors=errorRate    calculate link error rates for N seconds
  --testbeam            fixed IP address for testbeam
  --mspl=MSPL           Specify MSPL. Must be in the range 1-8 (default is 4)
  --nevts=nevts         Number of events to count at each scan point
[...]

PR #192, opened 2018-07-09
New Feature: configuration of all vfat params

bdorney:
## Description

Addresses #190.

### Types of changes

## Motivation and Context

User was unable to configure `latency` either using a command line argument or the `chamber_vfatDACSettings` dictionary for v2b electronics.  This PR enables `chamber_vfatDAQSettings` to set all `vfat2` parameters at runtime and also provides an optional command line argument (`-l` or `--latency`) for users who are not in the sudoers list (necessary to edit the `chamberInfo.py` file on system installations).

## How Has This Been Tested?

Full functionality of both `confChamber.py` and `confAllChambers.py` tested in:  http://cmsonline.cern.ch/cms-elog/1051618

## Checklist:


confChamber.py
```diff
@@ -43,20 +45,25 @@
...
-    parameters.defaultValues["IPreampFeed"] = chamber_vfatDACSettings[options.gtx]["IPreampFeed"]
-    parameters.defaultValues["IPreampOut"] = chamber_vfatDACSettings[options.gtx]["IPreampOut"]
-    parameters.defaultValues["IShaper"] = chamber_vfatDACSettings[options.gtx]["IShaper"]
-    parameters.defaultValues["IShaperFeed"] = chamber_vfatDACSettings[options.gtx]["IShaperFeed"]
-    parameters.defaultValues["IComp"] = chamber_vfatDACSettings[options.gtx]["IComp"]
+    for regName in chamber_vfatDACSettings[options.gtx].keys():
+        if regName not in parameters.defaultValues:
+            print("%s for link %d not recognized, please cross-check chamber_vfatDACSettings"%(regName, options.gtx))
+            print("list of possible values:", parameters.defaultValues.keys())
+            exit(os.EX_USAGE)
```
jsturdy, line 55: are you sure you want to exit here? rather than just print the message and continue?

confAllChambers.py
```diff
@@ -56,6 +59,8 @@ def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,cName,ztrim,debug=False):
 
     parser.add_option("--config", action="store_true", dest="config",
                       help="Set Configuration from simple txt files", metavar="config")
+    parser.add_option("-l","--latency", type="int", dest = "latency", default = None,
```
jsturdy: i would probably also remove the `-l` option, leaving just `--latency`, despite the fact that it may mean someone has to type a little more (i you care about typing, you could change it to `--lat`)

confChamber.py
bdorney: If a `$USER` has specified a `regName` in the `chamber_vfatDACSettings` and it is not found in `parameters.defaultValues` then there configuration won't be what they are expecting.  

So I don't think printing and continuing will do much good, e.g. their configuration will still be wrong.  Additionally it might be really wrong and if other arguments are specified (e.g. `--vfatConfig`) any error message here will be buried in further output.

I think it would be more reliable to have this throw an error and be the last message printed to terminal (additionally this should be caught by `confAllChambers.py` as a problem).

confAllChambers.py
bdorney: Okay, I can change this.

bdorney: Done

jsturdy (approved):

PR #193, opened 2018-07-14
New Feature: Correct trim algorithm implementing trim dac calibration [V3]

bdorney:
### New qcutility: `launchSCurve(...)`
Added a new utility function `launchSCurves(**kwargs)` which will launch an scurve using a defined set of parameters specified by the input arguments.  Most notably it can set the channel registers before launching an scurve.  Supported arguments are:

| Name | Type | Description |
| :----: | :----: | :----------- |
| `calSF` | `int` | value of the `CFG_CAL_FS` register, used in current injection mode of calibration pulse. |
| `cardName` | `str` | Name or ip address of AMC in uTCA crate |
| `chMask` | `int` array | Array of size 3072, indicates channels to mask.  Array index given by `idx = vfatN * 128 + channel`.|
| `chMax` | `int` | maximum channel to scan in scurve. |
| `chMin` | `int` | minimum channel to scan in scurve. |
| `debug` | `bool` | print debugging information. |
| `filename` | `str` | physical filename indicating absolute path of scurve outputfile. |
| `latency`  | `int` | Value of `CFG_LATENCY` to use when taking the scruve. | 
| `link` | `int` | optohybrid number on `cardName`. |
| `mspl` | `int` | value of `MSPL` or `CFG_PULSE_STRETCH` to use. |
| `nevts` | `int` | number of events to take in the scan. |
| `setChanRegs` | `bool` | write VFAT channel registers if `True`. |
| `vfatmask` | `int` | `vfatmask` to apply, 24-bit number, 1 in n^th bit indicates n^th vfat is masked. |
| `voltageStepPulse` | `bool` | Use voltage step pulse (`True`) instead of current pulse (`False`). |
| `trimARM` | `int` array | Array of size 3072, indicating trim amplitude to set for arming comparator. Array index given by `idx = vfatN * 128 + channel` |
| `trimARMPol` | `int` array | as `trimARM` but sets trim polarity. |
| `trimZCC` | `int` array | as `trimARM` but for the ZCC comparator. |
| `trimZCCPol` | `int` array | as `trimZCC` but sets trim polarity. |

### Updates to GEM Tree Structure
The tree structure has been changed to include additional types of trees and feature inheritance.

The base tree structure is now defined by the class `gemGenericTree` and declares all branches that are common to all tree types, namely:

- `link`
- `Nev`
- `utime`
- `ztrim`
- `vfatCH`
- `vfatID`
- `vfatN`

The base class also defines the methods common to all classes:

- `autoSave()`
- `getMode()`
- `setDefaults()`
- `write()`

Then several classes are defined that inherit from the base class:

- `gemDacCalTreeStructure`
- `gemTreeStructure`

The `gemDacCalTreeStructure` provides a `TTree` interface for storing the calibration of a `vfat2/3` DAC against some other DAC or value.  The `gemTreeStructure` class is the same class as is currently used.  Each class defines independently the relevant `TBranch`es within the `TTree`.

In each class the `fill(**kwargs)` method is defined separately (although a base method could be fined to take care of filling the common branches...this was not done though).

The change is backwards compatible and preserves all current information.

### Updates to `trimChamberV3.py`
The current trim algorithm is logically flawed and doesn't work.  The algorithm has been changed to implement the algorithm @aamir77i worked out.  It will now determine a set of channel register values that will, for a given `CFG_THR_ARM_DAC`, ensure that the scurve mean position of all channels on a vfat is aligned.

The proposed algorithm requires:

- Calibration of `CFG_CAL_DAC` to `fC` for each `vfat3`
- Calibration of `CFG_THR_ARM_DAC` to `fC` for each `vfat3` (see [this PR](https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/141) for more details).

These must be respectively supplied at runtime as a calibration files to the `--calFileCAL` and `--calFileARM` arguments.  Note the DAC settings used to generate the `--calFileARM` file must be the same settings that are used to take the trim run or else it will *not* work, specifically:

- `CFG_PT`
- `CFG_BIAS_SH_I_BDIFF`
- `CFG_BIAS_SH_I_BFCAS`
- `CFG_BIAS_SD_I_BDIFF`
- `CFG_BIAS_SD_I_BFCAS`
- `CFG_RES_PRE`
- `CFG_CAP_PRE`

See vfat3 manual for their meanings.

The trimming will take place at a `CFG_THR_ARM_DAC` value specified by the `--armDAC` argument for all VFATs.  A per VFAT value of `CFG_THR_ARM_DAC` can be specified by instead providing the `--vfatConfig` argument with a `vfatConfig.ch` file generated by either `anaUltraThreshold.py` or `anaSBitThresh.py`.

Then the trim algorithm will take an scurve for each of the trimDAC values specified in the `--trimPoints` argument.  The default uses the set `{-63, 0, +63}`, this is expected to be a comma separated list of integers.  Which covers the range of the trim amplitude register `[0x0,0x3F]` and the sign of positive and negative is included in the trim polarity bit (which determines if the amplitude is positive or negative).  For example:

trimDAC: -63
summary_vfat0_trimdac63_trimpol1

trimDAC: 0
summary_vfat0_trimdac0_trimpol0

trimDAC: +63
summary_vfat0_trimdac63_trimpol0

To prevent lost time and to provide the ability to **resume** a trim run the algorithm will first check to see if an scurve raw output file for the given trim point exists, is readable, and has the output `scruveTree` in the `--dirPath` location.  If the `--dirPath` argument has not been specified it will algorithmically default to:

```bash
$DATA_PATH/chamber_config[link]/trim
```

Where `chamber_config` is the standard mapping dictionary defined in [chamberInfo.py](https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/43f46bcc53fbcac83a34b3ccde94273c904d0395/mapping/chamberInfo.py#L7-L23).

Then for a given vfat channel the set of scurves will be fit and from each fit the scurve mean position in terms of fC will be extracted and plotted against the signed trim amplitude value the scurve was taken at.  This will then be fit with a `trimDac=m*charge+b` line, for example: 

trimdac_calibration_vfat0_chan1

This then parameterizes the trim amplitude register in terms of charge.  Once the trim dac calibration for each channel is known the trim amplitude and polarity values can be determined.  For this the following procedure is used for each VFAT:

1. The global arm dac value (e.g. `CFG_THR_ARM_DAC`) is converted into `fC` using the calibration defined by the `--calFileARM` value,
2. The trimDAC value to place this channel at that charge is determine via the `trimDAC = m*charge+b` equation determined above,
3. If trimDAC value is negative, trim polarity is set from `0x0` to `0x1` and then trimDAC is set its absolute value.
4. The trimDAC and trim polarity are then written to the channel of interest

After steps 1 through 4 are performed for all channels a final scurve is taken at the trimmed setting, the VFATs will now be trimmed:

summary_vfat0_trimmed

You can see that the scurve mean positions are aligned after trimming:
scurvemeanpos_vfat0_before_after_trimming

Here the red distribution is after trimming.  The blue distribution is before trimming.  The trimming procedure has correctly aligned the scurve mean position and trimmed the comparator such that all channels will have the same 50% response point.

The algorithm will generate a set of output files in the `--dirPath` directory (either user specified or generated):
- One scurve output file for each of the N values in `--trimPoints`
   - These will follow the naming convention: `SCurveData_trimdacX_trimPolY.root`
   - Here X is the absolute value of the N^th point
   - Here Y is 0 (1) if X is positive (negative). 
- One output file `TrimData.root` which stores:
   - The trimDAC calibration information in a `gemDacCalTreeStructure` instance called `trimDacArmTree`
   - The `CFG_CAL_DAC` calibration information in a `gemDacCalTreeStructure` instance called `calDacCalibration`
   - The `CFG_THR_ARM_DAC` calibration information in a `gemDacCalTreeStructure` instance called `thrArmDacCalibration`
   - A `TGraphErrors` object for each vfat channel specifying the trimDac calibration and a `TF1` object that is the fit.
- One `chConfig.txt` file that writes the trim information and can be passed to `confChamber.py` with the `--chConfig` argument to set the channel registers. 

The layout of the `chConfig.txt` file has been updated because `vfat3` is different from `vfat2`:
```
vfatN/I:vfatID/I:vfatCH/I:trimDAC/I:trimPolarity/I:mask/I
0       16777216        0       8       1       0
0       16777216        1       5       0       0
0       16777216        2       14      0       0
0       16777216        3       13      0       0
0       16777216        4       6       0       0
0       16777216        5       4       0       0
0       16777216        6       2       1       0
0       16777216        7       8       0       0
0       16777216        8       40      0       0
...
...
```

Note that channel masks will *not* be determined by `trimChamberV3.py`. 

### Types of changes

## Motivation and Context

V3 trim algorithm is not correct.

## How Has This Been Tested?

Extensively

Example call:
```bash
trimChamberV3.py -c eagle60 -g1 --vfatmask=0x18000 --armDAC=100 --latency=33 --mspl=4 --nevts=100 --voltageStepPulse --calFileCAL=scurve/calFile_GE11-X-S-CERN-0011_NoInvertConvention.txt --calFileARM=calFile_thrArmDAC_GE11-X-S-CERN-0011.txt
```

Before trimming:
summary

After trimming:
summary

## Checklist:


aamir77i:
Brian its Aamir here, good to see your results.
Could you calculate the ratio between threshold spread(Mean Threshold for one chip ) before trimming and compare with spread after trimming?
The number will show the effectiveness of the algorithm.

bdorney:
Hi @aamir77i that could be done, but I think it's not appropriate for this PR.  That might be more appropriate for a dedicated script in the https://github.com/cms-gem-daq-project/gem-plotting-tools repository.  Some "analyse trim data" script.

aamir77i:
Ok thanks

jsturdy (approved):

PR #196, opened 2018-07-26
New Feature: Readback Channel Registers in Python Tools [V3]

bdorney:
## Description

Goal to enable `readBackCheck()` for `confChamber.py` and allow `ultraScurve.py` and `ultraThreshold.py` to store channel register information when storing raw data.

Right now once channel registers are set they are never stored in any of the output data and cannot be ready back without a large number of atomic transactions.

Requires: https://github.com/cms-gem-daq-project/cmsgemos/pull/192

### Types of changes

## Motivation and Context

Provides functionality to cross-check that after writing configuration it matches expectation (by providing `readBackCheckV3(...)`) and stores this information in our "ultra" scans.

## How Has This Been Tested?

http://cmsonline.cern.ch/cms-elog/1055032

## Checklist:

### DIFF
--- a/confChamber.py
+++ b/confChamber.py
@@ -84,7 +84,7 @@ if __name__ == '__main__':
 
             dict_readBack = {}
             if vfatBoard.parentOH.parentAMC.fwVersion > 2:
-                dict_readBack = { "trimDAC":"VFAT_CHANNELS.CHANNEL", "trimPolarity":"VFAT_CHANNELS.CHANNEL", "mask":"VFAT_CHANNELS.CHANNEL" }
+                dict_readBack = { "trimDAC":"VFAT_CHANNELS.CHANNEL", "trimPolarity":"VFAT_CHANNELS.CHANNEL", "mask":"VFAT_CHANNELS.CHANNEL", "vfatID":"HW_CHIP_ID" }
                 print 'Comparing Currently Stored Channel Registers with %s'%options.chConfig
                 readBackCheckV3(chTree, dict_readBack, vfatBoard, options.vfatmask)
     
@@ -102,7 +102,7 @@ if __name__ == '__main__':
 
             dict_readBack = {}
             if vfatBoard.parentOH.parentAMC.fwVersion > 2:
-                dict_readBack = { "trimDAC":"VFAT_CHANNELS.CHANNEL", "trimPolarity":"VFAT_CHANNELS.CHANNEL", "mask":"VFAT_CHANNELS.CHANNEL" }
+                dict_readBack = { "trimDAC":"VFAT_CHANNELS.CHANNEL", "trimPolarity":"VFAT_CHANNELS.CHANNEL", "mask":"VFAT_CHANNELS.CHANNEL", "vfatID":"HW_CHIP_ID" }
                 print 'Comparing Currently Stored Channel Registers with %s'%options.chConfig
                 readBackCheckV3(chTree, dict_readBack, vfatBoard, options.vfatmask)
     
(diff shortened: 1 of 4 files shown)


jsturdy (dismissed):
No issues in a cursory review

jsturdy:
@bdorney, can you do the conflict resolution, I don't want to try as it seems to just be the entire file, so you should know best what should be the right one.

bdorney:
> @bdorney, can you do the conflict resolution, I don't want to try as it seems to just be the entire file, so you should know best what should be the right one.

I have resolved the conflicts.  You'll need to review/approve again.

confChamber.py
```diff
@@ -158,10 +144,10 @@
                     if not (vfatBoard.parentOH.parentAMC.fwVersion > 2):
                         vfatBoard.writeVFAT(int(event.vfatN), "ContReg3", int(event.trimRange),options.debug)
             
-            #print 'Comparing Curently Stored VFAT Registers with %s'%options.vfatConfig
-            #if options.vt1bump != 0:
-            #    print "Mismatches between write & readback valus for VThreshold1 should be exactly %i" %(options.vt1bump)
-            #readBackCheck(vfatTree, dict_readBack, ohboard, options.gtx)
+            if vfatBoard.parentOH.parentAMC.fwVersion > 2:
```
mexanick, line 147: this looks like independent of "compare" option. Why?

qcutilities.py
```diff
@@ -284,6 +333,137 @@ def readBackCheck(rootTree, dict_Names, device, gtx, vt1bump=0):
...
+    for bName,regName in dict_Names.iteritems():
+        print "Reading back (%s,%s) from all VFATs, any potential mismatches will be reported below"%(bName,regName)
+        print "="*40
+
+        regMap = []
+        regValues = []
+        if regName == "VFAT_CHANNELS.CHANNEL": #Channel Register
+            regValues = getChannelRegisters(vfatBoard,mask)
+            for vfat in range(0,24):
+                if (mask >> vfat) & 0x1:
```
mexanick, line 428: should we check here if the mask was applied correctly?

confChamber.py
bdorney: The script will always perform the comparison, independent of if the `--compare` option is supplied.  It will only write the registers if the comparison option is *not* supplied.  This maintains the same functionality of the v2b electronics, see for example:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/a5747b11e5f71a104a71c28132f97aef2996f2e2/confChamber.py#L119-L130

qcutilities.py
bdorney: I'm not sure I understand what you are referring to.

In L428 `mask` comes from the user supplied `vfatmask` and is not a writeable register; the channel mask is bit number 15 of the integers stored in `regValues` (L426) and is checked below at L435.

confChamber.py
mexanick: ah, ok

qcutilities.py
mexanick: we can provide a check to see whether this VFAT is in a sleep mode (if accessible) or inaccessible. So that we dont have unexpected extra chip in run mode with garbage values in its settings

bdorney: I'm not sure I understand.  This function can check the currently stored register values against some externally supplied values.  

The list of supported registers is defined in `list_KnownRegs`.  So some other script should call this function (`readBackCheckV3`) for say `CFG_RUN` or some other register of interest (provided a `TTree` with the stored values of `rootTree` is provided).  But to add such a check here is now how the function was designed to be used.

Right now we do not have a `TTree` which defines say `CFG_RUN`.

bdorney: Again the goal here with this PR is to continue to bring the existing `v2b` functionality to the `v3` tools.

jsturdy (approved):

mexanick (approved):

PR #208, opened 2018-10-21
Create anaSbitReadout Type: Enhancement

fraivone:
This script unpacks and plots data acquired with sbitReadout.py. The outputs are saved in `inputfolder/sbitReadOut/`
## IMPORTANT
1. the function getMapping in `gempython/gemplotting/utils/anautilities` needs to be modified. At the moment it works with VFAT2 map, in which channel number starts from 1. In order to use the script, I modified getMapping in my Personal Repository as suggested here https://mattermost.web.cern.ch/cms-gem-daq/pl/y6r6cqn1rbfa7m38iwn1hb9t5e
5. in the folder `gempython/gemplotting/mapping/` there must be located the files named `shortChannelMap_VFAT3-HV3b-V1_VFAT3-HV3b-V2.txt` `longChannelMap_VFAT3-HV3b-V1_VFAT3-HV3b-V2.txt`, which are the outputs of Louis's pull request https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/134/files#diff-5a657d72aa6d1b867b7000684423a4b1
### USAGE EXAMPLE
```bash
anaSbitReadout -i folder_containing_sbitReadout_outputs/ -t short
```
### FURTHER DEVELOPMENTS
1. Default GEBType is long, so the user may run the script without specifying it  while working with a short GEB. This will cause a bad string <-> channel mapping
2. The path to the .txt file with the channel <-> string mapping could be provided as input; it would be easier to use different mappings
3. An option to mask noisy channel or whole vfat could be useful
### KNOWN ISSUES
* While loading the .dat file, ROOT returns this warning: 
```
Warning in <TTree::ReadStream>: Couldn t read formatted data in "evtNum/i:sbitClusterData0/i:sbitClusterData1/i:sbitClusterData2/i:sbitClusterData3/i:sbitClusterData4/i:sbitClusterData5/i:sbitClusterData6/i:sbitClusterData7/i" for branch evtNum on line 1; ignoring line
``` 
I think it's cause ROOT doesn't expect to find headers in the .dat files opened with the ReadFile function.  I don't know how to solve this, however I don't think it's a big deal
* the parser is imported from `gempython/gemplotting/utils/anaoptions`, so it contains more options than actually implemented. This could be a problem for the user when reading the help menu

### Types of changes

Example of output in case of a Cd109 source placed underneath VFAT7
immagine

## Checklist:

### DIFF
--- a/anaSbitReadout
+++ b/anaSbitReadout
@@ -1,0 +1,375 @@
+#!/bin/env python
+"""
+    Script to analyze sbitreadout data
+    By: Caterian Aruta (caterina.aruta@cern.ch) and Francesco Ivone (francesco.ivone@cern.ch)
+    """
+if __name__ == '__main__':
+    import ROOT as r
+    import numpy as np
+    import root_numpy as rp
+    import glob
+    import sys
+    import os
+    import time
+    from optparse import OptionParser
+    from subprocess import call
+    from array import array
+    from gempython.utils.nesteddict import nesteddict as ndict
+    from gempython.gemplotting.mapping.chamberInfo import chamber_iEta2VFATPos
+    from gempython.gemplotting.mapping.chamberInfo import chamber_vfatPos2iEtaiPhi as vfat_to_etaphi
+    from gempython.gemplotting.utils.anautilities import make3x8Canvas, saveSummaryByiEta, getMapping
+    from gempython.gemplotting.utils.anaoptions import parser
+    import pkg_resources
+
+    parser.set_defaults(outfilename="sbitReadOut.root")
+    (options, args) = parser.parse_args()
+    path = options.filename
+    size = ((options.GEBtype).lower())
+
+    if not os.path.isdir(path):
+        raise OSError(2, "No such file or directory", path)
+    else:
+        pass
+
+    if size not in ('long', 'short'):
+        raise AssertionError("Invalid value of GEBtype")
+    else:
+        pass
+
+    filename = options.filename+"/sbitReadOut"
+    outfilename = options.outfilename
+    print("Analyzing: '%s'" % path)
+    os.system("mkdir "+filename)
+
+    """
+    At the moment the output of the sbitReadout.py are .dat files with headers that will automatically fill the ROOT TTree. However ther is one last ':' in the first line that shouldn't be there; consequently the ReadFile function is not able to understand the header. So before reading these .dat files, one has to be sure to remove the ':'
+    To achieve this the following command is needed (removes the last : in the header of all .dat files in the input path)
+    """
+    os.system("find "+path+" -iname \*.dat -exec sed -i 's/7\/i:/7\/i/g' {} \;")
+
+    # Set default histo behavior
+    r.TH1.SetDefaultSumw2(False)
+    r.gROOT.SetBatch(True)
+    r.gStyle.SetOptStat(1111111)
+
+    mappath = pkg_resources.resource_filename(
+        'gempython.gemplotting', 'mapping/')
+    # Loading the dictionary with the mapping
+    vfat_ch_strips = getMapping(
+        mappath+size+"ChannelMap_VFAT3-HV3b-V1_VFAT3-HV3b-V2.txt", False)
+    print("\nVFAT channels to strips "+size+":\t MAP loaded")
+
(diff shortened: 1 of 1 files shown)


bdorney:
@fraivone closing this PR, please re-submit this PR to [gem-plotting-tools](https://github.com/cms-gem-daq-project/gem-plotting-tools/pulls) (since this is a plotting tool).

anaSbitReadout
```diff
@@ -0,0 +1,375 @@
...
+    canv.Update()
+    canv.SaveAs(filename+'/ChvsiEta.png')
+
+    outF.Close()
+    print "\n---Took", (time.time() - start_time) / \
+        int(idx), "seconds for each .dat file---"
+    print "\nGaranting permission to ", filename, "..."
+    os.system("chmod -R 770 "+filename)
+    print "Data stored in", filename+'/'+outfilename, "\nBye now"
+    exit(-1)
```
jsturdy: Is this a good exit or a bad exit (there are standardized exit codes, so it's best to not try to reinvent the wheel)?
@bdorney enumerated some standard exit codes, and it would probably be good to use those here

anaSbitReadout
```diff
@@ -0,0 +1,375 @@
...
+    else:
+        pass
+
+    filename = options.filename+"/sbitReadOut"
+    outfilename = options.outfilename
+    print("Analyzing: '%s'" % path)
+    os.system("mkdir "+filename)
+
+    """
+    At the moment the output of the sbitReadout.py are .dat files with headers that will automatically fill the ROOT TTree. However ther is one last ':' in the first line that shouldn't be there; consequently the ReadFile function is not able to understand the header. So before reading these .dat files, one has to be sure to remove the ':'
```
jsturdy, line 45: Have you submitted a bug report to the appropriate location to have this fixed?
It can also be fixed by ignoring the header in the file and creating it in situ while analyzing (which will also fix the problem of the warning messages `ROOT` spits out when you loop over the files)

anaSbitReadout
```diff
@@ -0,0 +1,375 @@
+#!/bin/env python
+"""
+    Script to analyze sbitreadout data
+    By: Caterian Aruta (caterina.aruta@cern.ch) and Francesco Ivone (francesco.ivone@cern.ch)
+    """
+if __name__ == '__main__':
+    import ROOT as r
```
jsturdy, line 7: `ROOT` imports should be after parsing the options, otherwise there are instances where the `ROOT` option parser clobbers the option parser of the tool

anaSbitReadout
```diff
@@ -0,0 +1,375 @@
...
+        """
+    # Creating the output File and TTree
+    outF = r.TFile(filename+'/'+outfilename, 'recreate')
+    inT = r.TTree('Packed', 'Tree Holding packed raw data')
+
+    # searching for all the files with this format and adding them to the TTree
+    start_time = time.time()
+    print ("\nReading .dat files from the folder '%s'" % path)
+    for idx, file in enumerate(glob.glob(path+'/sbitReadOut_run*.dat')):
+        inT.ReadFile(file)
```
jsturdy, line 82: The `ROOT` warnings are happening because *each* of the input files has a header line, but only that of the *first* is used in creating the `TTree`.
This can be cleaned up by extracting the header from the first file, `cat`ing the rest of the files together and ignoring all header lines, and loading this `cat`ed file in with `ReadFile(catfile,headers)`

anaSbitReadout
jsturdy, line 375: Is this a good exit or a bad exit?
The exit codes that programs return are fairly standardized and have certain defined meanings, so it would be best to not try to do something new here.
Also, @bdorney has enumerated several common exit codes (somewhere) and those should probably be utilized here

jsturdy (commented):
Same comment as @bdorney re `gem-plotting-tools`, in addition to several comments on the PR itself

anaSbitReadout
fraivone: Ok, done

anaSbitReadout
fraivone: I've cated all the files together skipping the first lines, the header is inserted _by hand_. It works fine, no more warnings. Once the TTree is wirtten, the catfile is removed

PR #211, opened 2018-10-25
[V3] Update confAllChambers.py

bdorney:
## Description

In an attempt at a short term solution, and to play nice with xdaq, I've updated `confAllChambers.py` to work in v3 land.

Also `confAllChambers.py` will now:

 - Apply per link VFAT mask when using `chamber_vfatMask` dictionary,
 - And if `--config` option is supplied if a `chConfig` or `vfatConfig` file for a given link is not found it will print a warning and continue rather than exiting

### Types of changes

## Motivation and Context

Want a short term mechanism for configuring multiple links.  Would like that `GEMSupervisor` can configure the detectors with current python hack. 

## How Has This Been Tested?

```bash
(py2.7) [dorney@gem904qc8daq]~/scratch0/CMS_GEM/CMS_GEM_DAQ% confAllChambers.py --shelf=1 -s2 --config --run
No channel configuration exists for GE11-X-S-BARI-0010
No vfat configuration exists for GE11-X-S-BARI-0010
25 Oct 2018 16:08:58.510 [7f463ad04740] INFO - wrappers::runCommand <> - Executing command: confChamber.py --cardName=gem-shelf01-amc02 -g0 --run --vfatmask=0x40 --vt1bump=0 --vt1=100
2018.10.25.16.08
Open pickled address table if available  /opt/cmsgemos/etc/maps/amc_address_table_top.pickle...
Initializing AMC gem-shelf01-amc02
opened connection
Configuring VFATs with chamber_vfatDACSettings dictionary values
biased VFATs
Set CFG_THR_ARM_DAC to 100
VFATs set to run mode
Chamber Configured
[None]
Normal termination
```

## Checklist:


mexanick (approved):
Looks good after a quick glance

confAllChambers.py
```diff
@@ -64,11 +86,12 @@ def launchArgs(shelf,link,slot,run,vt1,vt1bump,config,cName,ztrim,debug=False):
                       help="VThreshold1 DAC value for all VFATs", metavar="vt1", default=100)
     parser.add_option("--vt1bump", type="int", dest="vt1bump",
                       help="VThreshold1 DAC bump value for all VFATs", metavar="vt1bump", default=0)
-
+    parser.add_option("--ztrim", type="float", dest="ztrim", default=4.0,
```
jsturdy, line 89: why was `ztrim` added here, but removed everywhere else?

bdorney: Short term hack to work with `xdaq`.  I see from `GEMSupervisor` that it is calling the `--ztrim` argument:

https://github.com/cms-gem-daq-project/cmsgemos/blob/1a982c60645af79652105acfc898ed7a6db8fd0f/gemsupervisor/src/common/GEMSupervisor.cc#L474-L493

I assumed these lines where uncommented when the rpm was built and this is what is being used (although I cannot check this since only the compiled libraries are installed on the machine). 

If `--ztrim` argument doesn't exist this will raise an error.  So I added it here to resolve this.

If you would prefer to remove the `--ztrim` from `GEMSupervisor` I could do that; but figured this was a step in the wrong direction.

jsturdy: no, that's fine, was just curious

jsturdy (approved):

Issue #206, opened 2018-10-16
Feature Request: Place `dacSelect` into output file made by `dacScanV3.py` Remove `isVFAT3`


bdorney:
## Brief summary of issue

Need to be able to distinguish in a dac scan `CFG_THR_ARM_DAC` as a bias current or bias voltage as discussed by @AndrewLevin and myself [here](https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/158#issuecomment-429650619).

Additionally we learned from Paul that the nominal bias current and voltage settings of `VFAT3a` are the same as `VFAT3b` but the range/size of the DAC's are different.  So there is no need to have a branch storing `isVFAT3a` in the output `TTree`.  This can be safely removed (as it is superfluous and only used here). 

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

## Expected Behavior

Add a `TBranch` to the output `TTree` produced by `dacScanV3.py` that stores `dacSelect`.

Remove the `isVFAT3a` `TBranch` from the output `TTree` produced by `dacScanV3.py`.

## Current Behavior

`dacSelect` is not stored `isVFAT3a` is stored.

## Context (for feature requests)

Described above.

bdorney:
Resolved by #215.

PR #215, opened 2018-11-14
dacScanV3.py now scans all DAC's by default

bdorney:
## Description

Addresses #206 by rebasing off of @AndrewLevin's branch.

I've changed the default behavior of `dacScanV3.py` and `run_scans.py` such that when a DAC scan is launched if the `--dacSelect` parameter is *not* provided all DAC's are scanned.  If `--dacSelect` is provided only that DAC will be scanned.  

### Types of changes

Breaking change since if `--dacSelect` is not provided it will not prompt the user for an input but just scan all DACs.

## Motivation and Context

Because it was really tedious to scan DAC's individually and produced *a lot* of scandate subdirectories which required tracking multiple files and multiple analyses.

In retrospect all DAC's being scanned should have been the *nominal* behavior from the beginning.

## How Has This Been Tested?

See: 
- http://cmsonline.cern.ch/cms-elog/1073862
- http://cmsonline.cern.ch/cms-elog/1073827

## Checklist:

### DIFF
--- a/dacScanV3.py
+++ b/dacScanV3.py
@@ -114,7 +114,6 @@ def scanAllLinks(args, calTree, vfatBoard):
                 vfatN = vfat
                 )
         if args.debug:
-
             print("| {0} | {1} | {2} | {3} | {4} | {5} | {6} | {7} | {8} | {9} |".format(
                 calTree.link[0],
                 calTree.vfatN[0],
--- a/dacScanV3.py
+++ b/dacScanV3.py
@@ -56,11 +56,13 @@ def scanAllLinks(args, calTree, vfatBoard):
         dacMin = dictDACInfo["dacMin"]
         calTree.nameX[0] = dictDACInfo["dacName"]
         dacSelect = dictDACInfo["dacSelect"]
+        calTree.dacSelect[0] = dacSelect
     else:
         dacMax = maxVfat3DACSize[args.dacSelect][0]
         dacMin = 0
         calTree.nameX[0] = maxVfat3DACSize[args.dacSelect][1]
         dacSelect = args.dacSelect
+        calTree.dacSelect[0] = dacSelect
 
     # Get VFAT register values
     from gempython.utils.nesteddict import nesteddict as ndict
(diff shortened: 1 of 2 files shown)


AndrewLevin:
This implies changes to the analysis script, which currently assumes there is one DAC register scanned:

https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/anaDACScan.py#L114

I guess I can rewrite it so that it will find all of the `dacSelect` values in the tree and then find the nominal values for each one, and the output text file will contain an extra column for the register name.

bdorney:
> This implies changes to the analysis script, which currently assumes there is one DAC register scanned:
> 
> https://github.com/cms-gem-daq-project/gem-plotting-tools/blob/develop/anaDACScan.py#L114
> 
> I guess I can rewrite it so that it will find all of the `dacSelect` values in the tree and then find the nominal values for each one, and the output text file will contain an extra column for the register name.

Yes I'm already working on it.  Try to focus on the PR's you already have open/issues assigned.

jsturdy:
> Addresses #206 by rebasing off of @AndrewLevin's branch.

Is there a reason this was done?

bdorney:
> > Addresses #206 by rebasing off of @AndrewLevin's branch.
> 
> Is there a reason this was done?

His PR was outstanding for over 20 days and it made sense to address #206 at the same time as these changes (@AndrewLevin already addressed #206 from what I saw).

jsturdy:
Sorry, the question was why the `rebase` if `AndrewLevin/for_issue_206` was not the original parent (which is not clear from the PR)
Doing it this way now means you're not starting from the tip of the target branch, and that if anything in the previous PR/commits was objected to, it's now included here.  In this case it doesn't look like there will be any issues, but we should probably avoid this in the future, and instead just bump the review of a stale PR.

jsturdy (dismissed):
seems ok

bdorney:
I should be scolded...testing was done with `dacScanV3.py` which succeeded...testing was not done with `run_scans.py` which [fails](https://mattermost.web.cern.ch/cms-gem-daq/pl/yzd9esozepdd98kt6npftdmngo).  Not following my own rules #Bad.  Do not merge.

bdorney:
Testing of `run_scans.py` shown here:

http://cmsonline.cern.ch/cms-elog/1074001

Test successful.

jsturdy (approved):

Issue #225, opened 2019-03-05
[Feature Request] Provided Option for testConnectivity.py  to write Found GBT Phase Settings to a File


bdorney:
## Brief summary of issue

We should provide two sets of output information:

 1. The results of the phase scan (Step 4) should be written to a file if an option `--writePhases2File` is provided.
 2. The found good phases should be written if VFATs are all properly synchronized (Step 5) if an option `--writePhases2File` is provided.
  - Note if the user skips step 4 and starts only from step 5 then the phases should not be written as the required dictionary isn't populated.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

### Writing Phase Scan Results to File During Step 4
If the option `--writePhases2File` is provided then you change:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/d38b9746bad1ccbdd4056c2406c15dd0055def7c/testConnectivity.py#L276-L278

Such that `outputFile` is not `None`, see documentation:

https://github.com/cms-gem-daq-project/xhal/blob/ea12e8928c6a3496ca6454a820e986c140ae4a7a/python/reg_interface_gem/core/gbt_utils_extended.py#L42-L51

If the `--writePhase2File` is not provided then `outputFile` should remain `None`.

### Writing Found Phases If Step 5 Is Successful
If the option flag `--writePhases2File` is provided for each OH in `ohMask` the phase scan results should be written to a file.  This should be done only if step 5 is passed, so after line 416 but before the DAC scan is attempted:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/d38b9746bad1ccbdd4056c2406c15dd0055def7c/testConnectivity.py#L415-L420

The good phase information is stored in the nested dictionary `dict_phases2Save`.

However if step 4 has been skipped (e.g. option `-f5` is provided) then the found good phases should not be written (since they will not be known), so after line 416 you should do something like:

```python
   if args.firstStep <= 4"
       for oh in range(nOHs):
            # skip masked OH's
            # write good phases in dict_phases2Save to file
```

## Current Behavior

The found GBT phases are not written to a file.

## Context (for feature requests)

We would need to know the good phases to store in the config DB for use with The Blaster (TM).

## Your Environment

* Version used: d38b9746bad1ccbdd4056c2406c15dd0055def7c

bdorney:
@BenjaminRS this is needed by mid march.

[...]

Issue #224, opened 2019-03-05
[Feature Request] Include Trigger Link Check in testConnectivity.py


bdorney:
## Brief summary of issue

Right now the connectivity testing is only checking the GBT paths between the backend and frontend electronics.  There is no check of the health of the trigger links.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

I think a step should be added between Step 3, Programming FPGA, and Step 4, GBT Phase Scans, that checks trigger link status is good.  An example of a good link status is shown [here](https://github.com/cms-gem-daq-project/sw_utils/blob/develop/v3ElectronicsUserGuide.md#checking-trigger-link-status).

This can be implemented by exposing [getmonTRIGGEROHmainLocal(...)](https://github.com/cms-gem-daq-project/ctp7_modules/blob/f6904b53c70f5189248d3fd0d02d558bec2ddfa6/src/daq_monitor.cpp#L84-L133) to the python API.  To do this, for the short term, an xhal interface to the remote callable version of the local function [getmonTRIGGEROHmain(...)](https://github.com/cms-gem-daq-project/ctp7_modules/blob/f6904b53c70f5189248d3fd0d02d558bec2ddfa6/src/daq_monitor.cpp#L135-L163) should be provided in the `legacyPreBoost` branch of `xhal`.  This would then need to be added to the the `HwAMC` class in `gempython`.

### Concerns for Only QC7

When working with detectors in some other test stand (e.g. coffin, QC8, sustained operation, etc...) we will only care about the GEM trigger link.

For QC7 we want to check that both the GEM trigger link (in the MTP12 connector panel) and the CSC trigger link (separate LC duplex connector on patch panel).  The way to do this would be for OHN to have the GEM trigger link going to OHN and the CSC trigger link going to OHN+1 on the fiber patch panel.  Then the code should check for a given OH that OHN and OHN+1 have the good status.  Here the provided `ohMask` with respect to the trigger should be treated as a even bits corresponding to OHN and odd bits corresponding to the CSC trigger link of OHN.

This behavior should only be checked if a given flag is provided, e.g. `--checkCSCTrigLink`.  Otherwise OHN should only correspond to OHN trigger link and the code should respect the given `ohMask`.

## Current Behavior

No check of the trigger link status is presently performed.

## Context (for feature requests)

Need a way to check the health of the trigger links before the superchamber is closed.

[...]

PR #232, opened 2019-03-20
Add link status check to test connectivity

BenjaminRS:
## Description

Adding the ability to check for the trigger link status, and also to write found GBT phase settings to a file.

### Types of changes

## Motivation and Context

This PR addresses #224 and #225.

## How Has This Been Tested?

Tested various different setups of the testConnectivity.py options where, for example, the link would fail.
## Checklist:

### DIFF
--- a/testConnectivity.py
+++ b/testConnectivity.py
@@ -71,6 +71,8 @@ def testConnectivity(args):
         args.cardName = getCardName(args.shelf,args.slot)
     if hasattr(args, 'chConfig') is False: # Text file containing channel configuration
         args.chConfig = None
+    if hasattr(args, 'checkCSCTrigLink') is False: # Using getTriggerLinkStatus with checkCSCTrigLink set to true
+        args.checkCSCTrigLink = False
     if hasattr(args, 'compare') is False: # Just Compare frontend settings?
         args.compare = False
     if hasattr(args, 'filename') is False: # TFile containing channel configuration
@@ -95,6 +97,8 @@ def testConnectivity(args):
         args.vfatConfig = False
     if hasattr(args, 'voltageStepPulse') is False: # Default to voltageStepPulse in Scurves
         args.voltageStepPulse = True
+    if hasattr(args, 'writePhases2File') is False: # Write found GBT Phase seetings to file
+        args.writePhases2File = False
     if hasattr(args, 'zeroChan') is False: # Zero all bits in all channel registers
         args.zeroChan = False
 
@@ -107,6 +111,7 @@ def testConnectivity(args):
     
     dataPath = os.getenv('DATA_PATH')
     gbtConfigPath = "{0}/OHv3c/20180717".format(os.getenv("GBT_SETTINGS")) # Ideally this would be a DB read...
+    elogPath = os.getenv('ELOG_PATH')
 
     # Initialize Hardware
     amc = getAMCObject(args.slot,args.shelf)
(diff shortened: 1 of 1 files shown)


testConnectivity.py
```diff
@@ -236,7 +239,17 @@ def testConnectivity(args):
             pass
 
         vfatBoard.parentOH.parentAMC.writeRegister("GEM_AMC.TTC.GENERATOR.ENABLE",0x0)
-        
+
+        print("Checking trigger link status:")
+        for trial in range(0,args.maxIter ):
+            testLinks = vfatBoard.parentOH.parentAMC.getTriggerLinkStatus(printSummary=True, checkCSCTrigLink=False, ohMask=args.ohMask)
```
bdorney: It seems this only partially addresses #224 since the `checkCSCTrigLink` argument to `HwAMC::getTriggerLinkStatus` is hardcoded to false.

I think we need to a command line option (see under `Concerns for Only QC7` under #224).

testConnectivity.py
```diff
@@ -236,7 +239,17 @@ def testConnectivity(args):
...
+	    if (testLinks < 1):
+                printGreen("Trigger link to OHs in mask:{0} is good".format(hex(args.ohMask)))
+                break
+            else:
+                printRed("Trigger link to OHs in mask:{0} failed".format(hex(args.ohMask)))
```
bdorney: I would change this to a `printYellow` and make the text more like:

> Trigger link to OHs in mask:{0} failed, retrying and issuing a reset to the trigger block of GEM_AMC

testConnectivity.py
```diff
@@ -236,7 +239,17 @@ def testConnectivity(args):
...
+                vfatBoard.parentOH.parentAMC.writeRegister("GEM_AMC.TRIGGER.CTRL.CNT_RESET",0x1)
```
bdorney: Ah I think this is the wrong reset (I probably told you the wrong register, my bad).  Here you should be using:

`GEM_AMC.TRIGGER.CTRL.MODULE_RESET`

This will reset the module, while above will just reset the counters (and cause them to roll up again if there's an issue).

testConnectivity.py
```diff
@@ -416,6 +433,15 @@ def testConnectivity(args):
...
         printGreen("VFAT Communication Successfully Established")
 
+    if args.writePhases2File and args.firstStep <= 4:
+        fName = elogPath+'/Phases.log'
+        fPhases = open(fName,"w") 
+        for ohN in range(nOHs):
+            # Skip masked OH's
+            if( not ((args.ohMask >> ohN) & 0x1)):
+                continue
+            fPhases.write("[{0}]: {1}".format(ohN,dict_phases2Save[ohN]))
```
bdorney: will this *always* write 24 numbers per OH?  e.g. what happens if it can't find a valid phase, I don't remember if I filled some dummy value to `dict_phases2Save[ohN]` then.

bdorney (changes requested):
Require option for checking CSC Trigger links and change of trigger reset register.

testConnectivity.py
BenjaminRS: Yes this currently writes 24 numbers per OH. e.g.:
[4]: [6, 2, 2, 4, 4, 2, 6, 2, 2, 2, 3, 3, 4, 2, 2, 4, 2, 2, 7, 5, 2, 2, 2, 2]
.. is there a way for me to test with no valid phase?

bdorney: simplest would be to insert a bad value into `dict_phases2Save[ohN]`.  Phase = 15 will always be bad.

BenjaminRS: ok checked with @bdorney -- the code wont actually execute if the phases were bad. Anyway updated this part to be ROOT friendly.

bdorney: Yeah no longer an concern.

bdorney (approved):

bdorney:
Looks good from my side.

jsturdy:
Closes #224 
Closes #225

PR #235, opened 2019-03-27
ignoreSyncErrs flag will also ignore phase finding errors

AndrewLevin:
Apply the `ignoreSyncErrs` flag to the GBT phase finding step.

## Description
Currently, the script `testConnectivity.py` will `return` if the proper phase is not found for any of the 24 VFATs, and there is no way to mask any VFATs. So, this pull request allows the user to ignore not found phases by setting the `ignoreSyncErrs` flag.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context
If some VFATs are missing from a detector, then the script will never find all 24 GBT phases and will return after the GBT phase finding step, which is not always desired.

## How Has This Been Tested?
Yes, I tested this on the coffin detector.

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #236, opened 2019-03-28
in testConnectivity.py, bias all VFATs 

AndrewLevin:
Bias all VFATs, even the ones that are not found in the database.

## Description
A call to `vfatBoard.biasAllVFATs` is made before the loop that writes `IREF` to all VFATs with calibration information.

### Types of changes

## Motivation and Context
Currently, VFATs that are not found in the database are never biased.

## How Has This Been Tested?
Yes, it was tested in combination with https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/235 on the coffin.

## Checklist:

### DIFF
--- a/testConnectivity.py
+++ b/testConnectivity.py
@@ -480,9 +480,12 @@ def testConnectivity(args):
             dict_vfat3CalInfo[ohN] = getVFAT3CalInfo(dict_chipIDs[ohN],debug=args.debug)
             if args.debug:
                 print("dict_vfat3CalInfo[{0}]:\n{1}".format(ohN,dict_vfat3CalInfo[ohN]))
-            
-            # Write IREF
+
+            #First, set the bias current for all VFATs to a standard value. Then, for each VFAT, check if that VFAT was found in the database. If it was found, then its bias current is replaced with the value from the database. If it was not found e.g. due to a bit flip in the chip ID, then the bias current is unchanged from the standard value.
             print("Setting CFG_IREF for all VFATs on OH{0}".format(ohN))
+            vfatBoard.biasAllVFATs(dict_vfatMask[ohN])    
+                
+
             for idx,vfat3CalInfo in dict_vfat3CalInfo[ohN].iterrows():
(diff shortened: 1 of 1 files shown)


testConnectivity.py
```diff
@@ -477,7 +477,10 @@ def testConnectivity(args):
             dict_vfat3CalInfo[ohN] = getVFAT3CalInfo(dict_chipIDs[ohN],debug=args.debug)
             if args.debug:
                 print("dict_vfat3CalInfo[{0}]:\n{1}".format(ohN,dict_vfat3CalInfo[ohN]))
-            
+
+            #VFATs found in the database get biased later. This ensures all VFATs get biased.
+            vfatBoard.biasAllVFATs(dict_vfatMask[ohN])
```
bdorney, line 486: The comment here is a bit vague as to the true nature of this line.  The goal here is to allow the user to have setup the VFAT configuration files to apply an IREF to *each* VFAT; then if a bit flip in the chipID results in a bad DB query the correct IREF value will still be applied.  Please change the comment to better reflect this.

Also squash your git commit history following directive from @jsturdy:

- [short instructions](https://github.com/cms-gem-daq-project/cmsgemos/pull/263#pullrequestreview-219547518)
- [detailed idea](https://github.com/cms-gem-daq-project/cmsgemos/pull/263#issuecomment-477534539)

An example of how to do this can be see [here](https://dev.to/blakedeboer/beginners-guide-to-interactive-rebasing-1ob)

AndrewLevin: done

bdorney (dismissed):

bdorney:
rebase needed

AndrewLevin:
rebased

testConnectivity.py
bdorney: In the future for a long comment like this consider using block comments, e.g.

```python
# this is a single line comment
"""
This is a block comment
It can be multi line
"""
```

bdorney (approved):

bdorney:
superceded by #245

Issue #248, opened 2019-04-12
[Feature Request] Update to Phase Finding Algorithm in testConnectivity.py


bdorney:
## Brief summary of issue

Currently phase set points chosen by `testConnectivity.py` may not be reliable.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

Paraphrasing an email from @evka85:

Phase 15 never works so we should just cut the scan range short (e.g. 0-14).

There should be 2 bad spots in the phase where the communication never works.  However the scan will sometimes only ever find one bad spot due to the finite resolution of the phase shifter.  In this case the bad spot is still there it's just between two values and being near it can cause problems.  The bad phase positions can also vary slightly with temperature by +/-1 position.

The bad phases are expected to be 8 phases apart (7 good values in between bad spots).  But this can vary by +/-1. For this reason the setpoint should never be further than 4 phase counts away from any bad spots.

So lines 417 to 491 should be re-written to find the dead spot(s).  If two dead spots are found then the algorithm should search between them to find the largest "good" window and use the middle phase of that window as the setpoint.  If one dead spot is found the algorithm should search 4 phases in either direction from the dead spot (cyclic boundary conditions) and see which direction yields the largest "good" window and then take the second to the last phase in the appropriate window as the phase set point.

## Current Behavior

The phase finding algorithm in `testConnectivity.py` is:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/1117886b61e0f2192c89db1313a6f7d74aa69889/testConnectivity.py#L417-L491

Here it loops through the phase scan results and looks to find a number of good phase points in a row.  The phase is considered good if the number of times the phase was good matches the number of attempts it was scanned.  The algorithm will then:

1. search for 5 good phases in a row, selecting the 3rd phase in the set as the phase to assign, 2. if 5 good phases in a row are not found it will search for 4 good phases in a row, selecting the 2nd phase in the set to assign as good,
3. if 4 good phases in a row are not found it will search for 3 good phases in a row, selecting the 2nd phase in a set to assign as good.
4. If 3 good phases in a row are not found it is declared a failure

## Context (for feature requests)

Bad phase setting of VFAT-GBT elinks can cause communication problems with lead to bus errors crashing calibration applications.

## Your Environment

* Version used:
* Shell used:

jsturdy:
[...]

PR #245, opened 2019-04-08
New Feature: Add dead channel by (OH,VFAT) List.

bdorney:
## Description

Production team requested the table that presents dead channels discovered during connectivity testing to be displayed for each VFAT on an OH rather than a sum over the OH.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context

Mentioned above.  Production team needs to know which VFAT may need to be replaced.

## How Has This Been Tested?

On QC7 setup.

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #249, opened 2019-04-17
Addresses #248, Updated Phase Finding Algorithm

bdorney:
## Description

Supersedes #245 and addresses issue #248 by including @jsturdy's updates to the phase finding algorithm.

Requires:

https://github.com/cms-gem-daq-project/ctp7_modules/pull/115
https://github.com/cms-gem-daq-project/xhal/pull/112
https://github.com/cms-gem-daq-project/cmsgemos/pull/273
https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/201

### Types of changes

Now the `shelf`, `slot` and `ohMask` arguments are required positional arguments.

## Motivation and Context

See #248 

## How Has This Been Tested?

On the QC7 stand with multiple hybrid configurations.

## Checklist:


testConnectivity.py
```diff
@@ -247,13 +250,14 @@ def testConnectivity(args):
         print("Checking GBT Communication (Before Programming GBTs)")
         if not gbtCommIsGood(vfatBoard.parentOH.parentAMC, doReset=True, printSummary=args.debug, ohMask=args.ohMask):
             printRed("Connectivity Testing Failed")
+            printYellow("If Vmon = 8.0V than Imon must be 1.71 +/- 0.01A; if not the GBT's are not locking to the fiber link")
```
jsturdy: `than` -> `then`
Is this a QC8 setup specific warning?

testConnectivity.py
```diff
@@ -418,66 +423,110 @@ def testConnectivity(args):
...
+                            phase2Write = int((badPhaseCounts[idx2Use[0]] - maxSeqPhase)/2+maxSeqPhase)
+                        else:
+                            phase2Write = int((minSeqPhase - badPhaseCounts[idx2Use[0]])/2+badPhaseCounts[idx2Use[0]])
+                            pass
                         pass
-                    pass # End loop over phases
+                    else:                       # Look for longest good window
+                        badPhaseCounts = np.sort(badPhaseCounts)
+                        ranges = []
+                        ranges.append(crange(0,
```
jsturdy: don't need `crange` here, as you're explicitly constructing ranges that will never wrap

bdorney: Why invent another function when an existing one will do?

testConnectivity.py
bdorney: No, the target voltage for post LS2 operation is 8.0V.

testConnectivity.py
jsturdy: using `range` is not inventing something ;-)

testConnectivity.py
jsturdy: I was more interested in the current draw, as we saw quite some variation in V2 current draw, so I was curious how well constrained this value is expected to be

testConnectivity.py
bdorney: Hah, okay fair point.  I'll implement that.

jsturdy: When reviewing the 3 bad phases steps, I was thinking that simply finding the distance between each pair would be simpler.

Say you have the following (taken from your real world example)
```
badPhases1 = [2,9,10]
badPhases2 = [2,8,10]
```
What phase would you like to choose?
In the current algorithm, you'd take `goodPhase1 = (9-2)/2+2`, and in the second you'd take `goodPhase2 = (8-4)` or `goodPhase2 = (10+4)`, right?
If you take the difference between successive elements, you'd end up with
```
deltaPhases1 = [7, 1, 8]
deltaPhases1 = [6, 2, 8]
```
If you then take the points that have the largest delta, and apply the case 1 algorithm (count both ways for max good phases), you'd wind up with `goodPhase1 = 10+4` and `goodPhase2 = 10+4` (though maybe one wouldn't want to take this phase, as it sits right next to 15 (are we choosing 14 as the phase for any VFATs that fall into the other cases currently?)

My motivation for this is that it's much simpler than the various cases

jsturdy: Here is what I used to test this:
```python
ar1 = np.array([2,9,10])
ar2 = np.array([2,8,10])
ar1p = np.array([ar1[(x+1)%len(ar1)] for x in range(len(ar1)) ])
ar2p = np.array([ar2[(x+1)%len(ar2)] for x in range(len(ar2)) ])
ar1res = (ar1p-ar1)%16
ar2res = (ar2p-ar2)%16
ar1res
ar2res
```

The middle two lines could be replaced with an appropriate `rotate` function, if such a function existed (I was not able to find one after a cursory search)

testConnectivity.py
bdorney: Ah I understand...yes in this case the current draw will probably depend on also the length of the LV cable powering the frontend.  I have not accounted for that.

jsturdy:
With the addition of each special corner case, is there any case (besides the "case 3: no bad phases") that can't be adequately chosen by the left/right best (maximum) window search?
Could set a minimum threshold, and if this is reached, break out of any loop over bad phase points
* 1 bad phase would choose correctly (by construction)
* 2 consecutive bad phases would choose correctly (effectively treating the two points as a "wide" case 1
* 2 non-sequential bad phases would be chosen correctly, regardless of their separation
* 3 bad phases would be chosen correctly
* 4+ bad phases?

From the example data in #248 this is the result:
```Ideal phase for  VFAT0 is:  [ 6] [200. 200.   0. 200. 200. 200. 200. 200. 200. 200.   0. 200. 200. 200. 200.   0.]
Ideal phase for  VFAT1 is:  [10] [200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0.   0.]
Ideal phase for  VFAT2 is:  [ 2] [200. 200. 200. 200. 200. 200.   0. 200. 200. 200. 200. 200. 200. 200.   0.   0.]
Ideal phase for  VFAT3 is:  [-1] [200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for  VFAT4 is:  [13] [200.   0. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for  VFAT5 is:  [ 4] [  0. 200. 200. 200. 200. 200. 200. 200.   0. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for  VFAT6 is:  [ 7] [200. 200. 200.   0. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for  VFAT7 is:  [ 7] [200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0. 200. 200. 200.   0.]
Ideal phase for  VFAT8 is:  [ 7] [200. 200. 200.   0. 200. 200. 200. 200. 200. 200. 200.   0. 200. 200. 200.   0.]
Ideal phase for  VFAT9 is:  [ 9] [200. 200. 200. 200. 200.   0. 200. 200. 200. 200. 200. 200. 200.   0. 200.   0.]
Ideal phase for VFAT10 is:  [ 4] [200. 200. 200. 200. 200. 200. 200. 200.   0. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for VFAT11 is:  [-1] [200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for VFAT12 is:  [ 5] [200.   0. 200. 200. 200. 200. 200. 200. 200.   0. 200. 200. 200. 200. 200.   0.]
Ideal phase for VFAT13 is:  [ 2] [200. 200. 200. 200. 200. 200. 173. 200. 200. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for VFAT14 is:  [-1] [200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for VFAT15 is:  [ 5] [200.   0. 200. 200. 200. 200. 200. 200. 200.   0. 200. 200. 200. 200. 200.   0.]
Ideal phase for VFAT16 is:  [-1] [200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for VFAT17 is:  [ 9] [200. 200. 200. 200. 200. 178. 200. 200. 200. 200. 200. 200. 200.  66. 200.   0.]
Ideal phase for VFAT18 is:  [ 8] [200. 200. 200. 200.   0. 200. 200. 200. 200. 200. 200. 200. 188. 200. 200.   0.]
Ideal phase for VFAT19 is:  [14] [200. 200.   0. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for VFAT20 is:  [10] [200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0.   0.]
Ideal phase for VFAT21 is:  [ 9] [200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200. 200.   0. 200.   0.]
Ideal phase for VFAT22 is:  [ 4] [  0. 200. 200. 200. 200. 200. 200. 200.   1. 200. 200. 200. 200. 200. 200.   0.]
Ideal phase for VFAT23 is:  [ 7] [200. 200. 200.   0. 200. 200. 200. 200. 200. 200. 200.   0. 200. 200. 200.   0.]
```
The only thing this doesn't take into account is proximity to 15, and any preference for directionality in the choice (i.e., prefer a phase that doesn't require a wraparound if possible)

bdorney:
I think the only way to know which is better is to perform testing on multiple detectors to ensure this reliably sets the phase for slow control; I think we don't have this ability and there would be push back from the production team and the higher ups.

While I understand the request for simplicity we *need* something that's working now.  If we want to improve this I would suggest opening issue and returning to this once we have 10-30 detectors on the QC8 stand.

Then we can see with high statistics which way is better.  But for now I'm adverse to changing the algorithm because we are behind and it is "working"

utils/phaseUtils.py
```diff
@@ -0,0 +1,110 @@
...
+
+    vfatBoard - Instance of HwVFAT class
+    vfat      - Number indicating vfat position in range [0,23]
+    phase     - Phase value to write in range [0,15]
+    """
+    
+    # Get hardware info
+    cardName = vfatBoard.parentOH.parentAMC.name
+    ohN      = vfatBoard.parentOH.link
+    phase    = vfatBoard.parentOH.vfatGBTPhases[vfat]
```
jsturdy, line 95: Is it by design that you overwrite the phase value passed in as an argument to the function?

bdorney: hah no.  Will need to fix that.

jsturdy:
Discussed with @bdorney, and as these new cases are showing up as failures of the current algorithm, the preferred route will be to implement the simpler algorithm.
He will not have time to do so in the near future, so someone can take the current work and make the change, we can get it merged in.
If I have time, I will do it within the week

bdorney:
Updated to be even with #250

bdorney:
superceded by #251

Issue #259, opened 2019-05-23
[Bug Report] run_scans.py does not discovers vfatConfig if using trim option


bdorney:
## Brief summary of issue

Title.

Otherwise trimming will either be done at ARM DAC = 100 or apply a single VFAT config to all links (bad).

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

Like it discovers calibration files it should discover VFAT Config.

## Current Behavior

It does not discover VFAT config.

## Context (for feature requests)

Trimming at wrong ARM DAC vs. used ARM DAC yields bad results, example:

http://cmsonline.cern.ch/cms-elog/1084287

## Your Environment

* Version used:
* Shell used:

bdorney:
# Solution to the Bug

The following lines:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/54711c0191d20c2c3cfa026b56d01eb6a393f07c/run_scans.py#L331-L332

Should be changed to something like:

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/54711c0191d20c2c3cfa026b56d01eb6a393f07c/run_scans.py#L282-L303

Simple implementation:

```python
        # Check to see if a vfatConfig exists
        vfatConfigFile = "{0}/configs/vfatConfig_{1}.txt".format(dataPath,chamber_config[ohKey])
        if os.path.isfile(vfatConfigFile):
            cmd.append("--vfatConfig={}".format(vfatConfigFile))
```

Then I would suggest to drop the `vfatConfig` argument for this sub command.

bdorney:
Addressed by #260

PR #260, opened 2019-06-06
Addressing Issue #259 and testConnectivity.py Detects Bad VFAT DAC Cases

bdorney:
## Description

Addressing #259.

Additionally, `testConnectivity.py` will now detect cases where a VFAT DAC cannot reach the right bias voltage/current setting.  An option `--acceptBadDACBiases` has been provided to be able to continue even when the DACs are found to be problematic.

Requires:

 - https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/215

### Types of changes

The `--vfatConfig` argument has been dropped from the `trim` subcommand help menu.

Now if the user does not specify the `--armDac` option the algorithm will look under `$DATA_PATH/configs` for each unmasked OH for a file of the form `vfatConfig_<detSN>.txt`-where detSN comes from `chamber_config.values()`-exists; if it does exist this file will be passed to `trimChamberV3.py` when the call is made so that a per `vfatConfig` can be applied at the time of trimming automatically.

## Motivation and Context

See #259 

## How Has This Been Tested?

Tested under  http://cmsonline.cern.ch/cms-elog/1085008

## Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/run_scans.py
+++ b/run_scans.py
@@ -328,8 +328,13 @@ def trimChamberV3(args):
         # Additional optional arguments
         if args.armDAC is not None:
             cmd.append("--armDAC={}".format(args.armDAC))
-        if args.vfatConfig is not None:
-            cmd.append("--vfatConfig={}".format(args.vfatConfig))
+        else:
+            # Check to see if a vfatConfig exists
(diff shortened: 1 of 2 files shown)


mexanick (approved):
Cursory

jsturdy (approved):
Cursory

PR #264, opened 2019-06-21
Bug Fix: Corrected potential ImportError

bdorney:
## Description

Noticed that `maxVfat3DACSize` was being imported from `amc_user_functions_xhal` and this was not raising an `ImportError`.  This dictionary is defined in `gempython.tools.hw_constants`, but this dictionary *is* imported from `hw_constants` in `amc_user_functions_xhal` so I guess this is why this works currently; but it would be more correct to import this from the source, e.g. `hw_constants`.

https://github.com/cms-gem-daq-project/cmsgemos/blob/generic-amc-RPC-v3-short-term/gempython/tools/amc_user_functions_xhal.py#L2

https://github.com/cms-gem-daq-project/cmsgemos/blob/48e2f269dac0947be0101d2a9f0d943752e2ce61/gempython/tools/hw_constants.py#L18-L44

### Types of changes

## Motivation and Context

Maintenance?

## How Has This Been Tested?

Change was trivial.

## Checklist:

### DIFF
--- a/dacScanV3.py
+++ b/dacScanV3.py
@@ -1,5 +1,5 @@
 #!/bin/env python
-from gempython.tools.amc_user_functions_xhal import maxVfat3DACSize
+from gempython.tools.hw_constants import maxVfat3DACSize
 
 import os
 


bdorney:
Indeed in python on the DAQ machine:

```python
>>> from gempython.tools.amc_user_functions_xhal import maxVfat3DACSize
>>>
>>> from gempython.tools.hw_constants import maxVfat3DACSize
>>>
```

Both imports work; but the second one I would view as "correct" since the dictionary is defined there.

jsturdy (approved):

PR #265, opened 2019-06-25
[new feature] GE2/1 compatibility

mexanick:
Implements required changes in the scan utils for GE2/1 compatibility.
Requires https://github.com/cms-gem-daq-project/cmsgemos/pull/287
### Types of changes

## How Has This Been Tested?

Tested at GE2/1 integration test stand @904
## Checklist:

### DIFF
--- a/confChamber.py
+++ b/confChamber.py
@@ -10,6 +10,7 @@ if __name__ == '__main__':
 
     from array import array
     from gempython.tools.vfat_user_functions_xhal import *
+    from gempython.tools.hw_constants import gemVariants
     from gempython.vfatqc.utils.qcutilities import getCardName, inputOptionsValid
     from gempython.vfatqc.utils.confUtils import configure, readBackCheckV3, setChannelRegisters
 
--- a/confChamber.py
+++ b/confChamber.py
@@ -7,7 +7,7 @@ if __name__ == '__main__':
     Modified by: Eklavya Sarkar eklavya.sarkar@cern.ch
                  Brian Dorney brian.l.dorney@cern.ch
     """
-    
+
     from array import array
     from gempython.tools.vfat_user_functions_xhal import *
     from gempython.vfatqc.utils.qcutilities import getCardName, inputOptionsValid
@@ -16,7 +16,7 @@ if __name__ == '__main__':
     import argparse
     from reg_utils.reg_interface.common.reg_xml_parser import parseInt
     parser = argparse.ArgumentParser(description="Tool for configuring front-end electronics")
-    
+
     parser.add_argument("--applyMasks", action="store_true", help="Channel masks defined in chConfig will be applied if this option is provided; otherwise no channel masks will be set")
     parser.add_argument("--chConfig", type=str, dest="chConfig", default=None,
                       help="Specify file containing channel settings from anaUltraSCurve.py")
@@ -48,17 +48,19 @@ if __name__ == '__main__':
                       help="VThreshold1 DAC bump value for all VFATs", default=0)
     parser.add_argument("--zeroChan", action="store_true", dest="zeroChan",
                       help="Zero all channel registers")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
     args = parser.parse_args()
-    
+
     import subprocess,datetime
     startTime = datetime.datetime.now().strftime("%Y.%m.%d.%H.%M")
     print startTime
     Date = startTime
-    
+
     cardName = getCardName(args.shelf,args.slot)
-    vfatBoard = HwVFAT(cardName, args.gtx, args.debug)
+    vfatBoard = HwVFAT(cardName, link=args.gtx, debug=args.debug, gemType=args.gemType, detType=args.detType)
     print('opened connection')
-    
+
     # Check args
     if not inputOptionsValid(args, vfatBoard.parentOH.parentAMC.fwVersion):
         print("input options are not valid, exiting")
--- a/dacScanV3.py
+++ b/dacScanV3.py
@@ -84,7 +84,7 @@ if __name__ == '__main__':
     parser.add_argument("shelf", type=int, help="uTCA shelf to access")
     parser.add_argument("slot", type=int,help="slot in the uTCA of the AMC you are connceting too")
     parser.add_argument("ohMask", type=parseInt, help="ohMask to apply, a 1 in the n^th bit indicates the n^th OH should be considered", metavar="ohMask")
-    
+
     # Optional arguments
     parser.add_argument("-d","--debug", action="store_true", dest="debug",
             help = "Print additional debugging information")
@@ -96,10 +96,12 @@ if __name__ == '__main__':
             help = "Specify output filename to store data in.")
     parser.add_argument("--series", action="store_true", dest="series",
             help = "Scan nonzero links in ohMask in series (successive RPC calls) instead of in parallel (one RPC call)")
-    parser.add_argument("--stepSize", type=int, dest="stepSize",default=1, 
+    parser.add_argument("--stepSize", type=int, dest="stepSize",default=1,
                   help="Supply a step size for the scan")
     parser.add_argument("-v","--vfatmask",type=parseInt,dest="vfatmask",default=None,
             help="VFATs to be masked in scan & analysis applications (e.g. 0xFFFFF masks all VFATs)")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
     args = parser.parse_args()
 
     from gempython.utils.gemlogger import printRed
@@ -114,28 +116,28 @@ if __name__ == '__main__':
     from gempython.vfatqc.utils.qcutilities import getCardName, inputOptionsValid
     cardName = getCardName(args.shelf,args.slot)
     from gempython.tools.vfat_user_functions_xhal import *
-    vfatBoard = HwVFAT(cardName, 0, args.debug) # Assign link 0; we will update later
+    vfatBoard = HwVFAT(cardName, 0, args.debug, args.gemType, args.detType) # Assign link 0; we will update later
     print 'opened connection'
     amcBoard = vfatBoard.parentOH.parentAMC
     if amcBoard.fwVersion < 3:
         printRed("DAC Scan of v2b electronics is not supported, exiting!!!")
         exit(os.EX_USAGE)
-    
+
     # Check options
     if not inputOptionsValid(args, amcBoard.fwVersion):
         exit(os.EX_USAGE)
         pass
-    
+
     # Make output files
     import ROOT as r
     outF = r.TFile(args.filename,"RECREATE")
-    from gempython.vfatqc.utils.scanUtils import dacScanAllLinks, dacScanSingleLink 
+    from gempython.vfatqc.utils.scanUtils import dacScanAllLinks, dacScanSingleLink
     from gempython.vfatqc.utils.treeStructure import gemDacCalTreeStructure
     calTree = gemDacCalTreeStructure(
                     name="dacScanTree",
                     nameX="dummy", # temporary name, will be over-ridden
                     nameY=("ADC1" if args.extRefADC else "ADC0"),
-                    dacSelect=-1, #temporary value, will be over-ridden 
+                    dacSelect=-1, #temporary value, will be over-ridden
                     description="GEM DAC Calibration of VFAT3 DAC"
             )
 
--- a/dacScanV3.py
+++ b/dacScanV3.py
@@ -86,6 +86,7 @@ if __name__ == '__main__':
     parser.add_argument("ohMask", type=parseInt, help="ohMask to apply, a 1 in the n^th bit indicates the n^th OH should be considered", metavar="ohMask")
 
     # Optional arguments
+    from gempython.tools.hw_constants import gemVariants
     parser.add_argument("-d","--debug", action="store_true", dest="debug",
             help = "Print additional debugging information")
     parser.add_argument("--dacSelect", type=int, dest="dacSelect",
--- a/getCalInfoFromDB.py
+++ b/getCalInfoFromDB.py
@@ -8,7 +8,9 @@ if __name__ == '__main__':
     parser.add_argument("link",type=int,help="OH on AMC slot")
     parser.add_argument("-d","--debug",action="store_true",help="Prints additional information")
     parser.add_argument("--write2File",action="store_true",help="If Provided data will be written to appropriate calibration files")
-    parser.add_argument("--write2CTP7",action="store_true",help="If Provided VREF_ADC and IREF data will be sent to the VFAT3 config files on the CTP7 (implies --write2File)")
+    parser.add_argument("--write2CTP7",action="store_true",help="If Provided IREF data will be sent to the VFAT3 config files on the CTP7")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default="short")
     args = parser.parse_args()
 
     # Update the files on the DAQ machine whenever an update on the CTP7 is requested
@@ -17,15 +19,15 @@ if __name__ == '__main__':
         args.write2File = True
 
     from gempython.tools.vfat_user_functions_xhal import *
-    
+
     from gempython.vfatqc.utils.qcutilities import getCardName
     cardName = getCardName(args.shelf,args.slot)
-    vfatBoard = HwVFAT(cardName,args.link)
+    vfatBoard = HwVFAT(cardName, link=args.link, gemType=args.gemType, detType=args.detType)
 
     mask = vfatBoard.parentOH.getVFATMask()
     chipIDs = vfatBoard.getAllChipIDs(mask)
 
-    while(len(chipIDs) != 24):
+    while(len(chipIDs) < vfatBoard.parentOH.nVFATs):
         chipIDs.append(0)
 
     if args.debug:
@@ -115,7 +117,7 @@ if __name__ == '__main__':
         import logging
         gemlogger = getGEMLogger(__name__)
         gemlogger.setLevel(logging.INFO)
-    
+
         from gempython.vfatqc.utils.confUtils import updateVFAT3ConfFilesOnAMC
         updateVFAT3ConfFilesOnAMC(cardName,args.link,filename_iref,"CFG_IREF")
         updateVFAT3ConfFilesOnAMC(cardName,args.link,filename_vref_adc,"CFG_VREF_ADC")
--- a/getCalInfoFromDB.py
+++ b/getCalInfoFromDB.py
@@ -2,6 +2,7 @@
 
 if __name__ == '__main__':
     import argparse
+    from gempython.tools.hw_constants import gemVariants
     parser = argparse.ArgumentParser(description="Queries the DB for calibration info for all VFATs on given (shelf,slot,link)")
     parser.add_argument("shelf",type=int,help="uTCA shelf number")
     parser.add_argument("slot",type=int,help="AMC slot in uTCA shelf")
--- a/getCalInfoFromDB.py
+++ b/getCalInfoFromDB.py
@@ -9,7 +9,7 @@ if __name__ == '__main__':
     parser.add_argument("link",type=int,help="OH on AMC slot")
     parser.add_argument("-d","--debug",action="store_true",help="Prints additional information")
     parser.add_argument("--write2File",action="store_true",help="If Provided data will be written to appropriate calibration files")
-    parser.add_argument("--write2CTP7",action="store_true",help="If Provided IREF data will be sent to the VFAT3 config files on the CTP7")
+    parser.add_argument("--write2CTP7",action="store_true",help="If Provided VREF_ADC and IREF data will be sent to the VFAT3 config files on the CTP7 (implies --write2File)")
     parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
     parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default="short")
     args = parser.parse_args()
--- a/monitorTemperatures.py
+++ b/monitorTemperatures.py
@@ -26,7 +26,7 @@ Positional arguments
     uTCA crate shelf number
 
 .. option:: slot
-   
+
     AMC slot number in the uTCA crate
 
 .. option:: ohMask
@@ -101,7 +101,7 @@ if __name__ == '__main__':
     parser.add_argument("-t","--timeInterval", type=int, dest="timeInterval", default=60,
             help ="Time interval, in seconds, to wait in between temperature reads",metavar="timeInterval")
     parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
-    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default="short")
     args = parser.parse_args()
 
     if args.filename is not None:
--- a/monitorTemperatures.py
+++ b/monitorTemperatures.py
@@ -99,6 +99,8 @@ if __name__ == '__main__':
             help = "Do not print VFAT temperatures to terminal; if --filename isp provided VFAT temperatures are still read and stored in the output file")
     parser.add_argument("-t","--timeInterval", type=int, dest="timeInterval", default=60,
             help ="Time interval, in seconds, to wait in between temperature reads",metavar="timeInterval")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
     args = parser.parse_args()
 
     if args.filename is not None:
@@ -113,7 +115,7 @@ if __name__ == '__main__':
     from gempython.tools.vfat_user_functions_xhal import *
     from gempython.vfatqc.utils.qcutilities import getCardName
     cardName = getCardName(args.shelf,args.slot)
-    vfatBoard = HwVFAT(cardName, 0, args.debug) # Set a dummy link for now
+    vfatBoard = HwVFAT(cardName, 0, args.debug, args.gemType, args.detType) # Set a dummy link for now
     amcBoard = vfatBoard.parentOH.parentAMC
     print('opened connection')
 
@@ -148,8 +150,8 @@ if __name__ == '__main__':
 
     from time import sleep, time
     from ctypes import *
-    adcDataIntRefMultiLinks = (c_uint32 * (24 * 12))()
-    adcDataExtRefMultiLinks = (c_uint32 * (24 * 12))()
+    adcDataIntRefMultiLinks = (c_uint32 * (vfatBoard.parentOH.nVFATs * 12))()
+    adcDataExtRefMultiLinks = (c_uint32 * (vfatBoard.parentOH.nVFATs * 12))()
     try:
         print("Reading and recording temperature data, press Ctrl+C to stop")
 
@@ -181,8 +183,8 @@ if __name__ == '__main__':
                 for ohN in range(0,12):
                     if( not ((args.ohMask >> ohN) & 0x1)):
                         continue
-                    for vfat in range(0,24):
-                        idx = ohN * 24 + vfat
+                    for vfat in range(0,vfatBoard.parentOH.nVFATs):
+                        idx = ohN * vfatBoard.parser.nVFATs + vfat
                         print("| {0} | {1} | {2} | {3} | {4} |".format(
                             ohN,
                             vfat,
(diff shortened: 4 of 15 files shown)


testConnectivity.py
```diff
@@ -242,22 +243,34 @@ def testConnectivity(args):
     envCheck('DATA_PATH')
     envCheck('ELOG_PATH')
     envCheck("GBT_SETTINGS")
-    
+
     dataPath = os.getenv('DATA_PATH')
-    gbtConfigPath = "{0}/OHv3c/".format(os.getenv("GBT_SETTINGS")) # Ideally this would be a DB read...
+    #if ars.gemType == "ge21":
+    #    gbtConfigPath = ""
+    #else:
```
bdorney: Delete commented lines here

testConnectivity.py
```diff
@@ -242,22 +243,34 @@ def testConnectivity(args):
...
+    if args.gemType == "ge11":
+        gbtConfigPath = "{0}/OHv3c/".format(os.getenv("GBT_SETTINGS")) # Ideally this would be a DB read...
+    elif args.gemType == "ge21":
+        gbtConfigPath = "{0}/OHGE21/".format(os.getenv("GBT_SETTINGS"))
+    else:
+        print("me0 gemType not currently implemented, exiting.")
+        printRed("Connectvity Testing Failed")
+        return
+
+
```
bdorney: Unnecessary white space

testConnectivity.py
```diff
@@ -275,25 +288,35 @@ def testConnectivity(args):
...
+        if args.gemType == "ge11":
+            gbtConfigs = [
                 "{0}/GBTX_OHv3c_GBT_0.txt".format(gbtConfigPath),
                 "{0}/GBTX_OHv3c_GBT_1.txt".format(gbtConfigPath),
                 "{0}/GBTX_OHv3c_GBT_2.txt".format(gbtConfigPath),
                 ]
+        elif args.gemType == "ge21":
+            gbtConfigs = [
+                "{0}/GBTX_GE21_OHv1_GBT_0_minimal_2019-01-24.txt".format(gbtConfigPath),
+                "{0}/GBTX_GE21_OHv1_GBT_1_minimal_2019-01-24.txt".format(gbtConfigPath),
```
bdorney: I would suggest to have these be generic symlinks rather then specific files as you will need to update this over time (based on GE1/1 experience)

testConnectivity.py
```diff
@@ -275,25 +288,35 @@ def testConnectivity(args):
...
+        else:
+            print("me0 gemType not currently implemented, exiting.")
+            exit(os.EX_USAGE)
+
         print("Programming GBTs")
         configGBT(cardName=args.cardName, listOfconfigFiles=gbtConfigs, ohMask=args.ohMask, nOHs=nOHs)
 
         print("Checking GBT Communication (After Programming GBTs)")
-        if not gbtCommIsGood(vfatBoard.parentOH.parentAMC, doReset=True, printSummary=args.debug, ohMask=args.ohMask):
+        if not gbtCommIsGood(vfatBoard.parentOH.parentAMC, doReset=True, printSummary=args.debug, ohMask=args.ohMask, gemType="ge11"):
```
bdorney: I think you want `args.gemType` here

testConnectivity.py
```diff
@@ -1196,10 +1227,10 @@ def testConnectivity(args):
...
 
     # Enforce a minimum number of phase scans
-    if args.nPhaseScans < 50:
-        printYellow("You've requested the number of phase scans to be {0} which is less than 50.\nThis is probably not reliable, reseting to 50".format(args.nPhaseScans))
-        args.nPhaseScans = 50
-        pass
+    #if args.nPhaseScans < 50:
+    #    printYellow("You've requested the number of phase scans to be {0} which is less than 50.\nThis is probably not reliable, reseting to 50".format(args.nPhaseScans))
+    #    args.nPhaseScans = 50
+    #    pass
```
jsturdy: if, as in @bdorney's comment above, we're removing commented lines, here's a candidate

bdorney:
I didn't quite finish my review of this PR; I'll try to do this by 17h00 today.

testConnectivity.py
```diff
@@ -2,11 +2,12 @@
 
 from gempython.tools.amc_user_functions_uhal import *
 from gempython.tools.amc_user_functions_xhal import NoUnmaskedOHException
-from gempython.tools.hw_constants import gemVariants, GBT_PHASE_RANGE 
+from gempython.tools.hw_constants import gemVariants, GBT_PHASE_RANGE, vfatsPerGemVariant
 from gempython.tools.optohybrid_user_functions_xhal import OHRPCException, OHTypeException
 from gempython.tools.vfat_user_functions_xhal import *
 from gempython.utils.gemlogger import colors, getGEMLogger, printGreen, printRed, printYellow
-    
+#from gempython.utils.wrappers import runCommand
```
bdorney: delete line

testConnectivity.py
```diff
@@ -745,17 +768,22 @@ def testConnectivity(args):
                         name="dacScanTree",
                         nameX="dummy", # temporary name, will be over-ridden
                         nameY=("ADC1" if args.extRefADC else "ADC0"),
-                        dacSelect=-1, #temporary value, will be over-ridden 
+                        dacSelect=-1, #temporary value, will be over-ridden
                         description="GEM DAC Calibration of VFAT3 DAC"
                 )
- 
+
+        printYellow("CAlTree retrieved")
```
bdorney, line 777: Seems like this is a debugging statements that are added but no longer needed....?

testConnectivity.py
```diff
@@ -745,17 +768,22 @@ def testConnectivity(args):
...
         # Place All Chips Into Run Mode and write correct Iref
         from math import isnan
+        printYellow("isnan imported")
```
bdorney: Seems like this is a debugging statements that are added but no longer needed....?

testConnectivity.py
```diff
@@ -745,17 +768,22 @@ def testConnectivity(args):
...
         for ohN in range(nOHs):
+            printRed("Checking whether OH%s is masked" %(ohN))
```
bdorney, line 781: Seems like this is a debugging statements that are added but no longer needed....?

testConnectivity.py
```diff
@@ -745,17 +768,22 @@ def testConnectivity(args):
...
             # Skip masked OH's
             if( not ((args.ohMask >> ohN) & 0x1)):
                 continue
 
+            printRed("Working with OH%s" %(ohN))
```
bdorney: Seems like this is a debugging statements that are added but no longer needed....?

utils/scanUtils.py
```diff
@@ -16,6 +16,7 @@ def dacScanAllLinks(args, calTree, vfatBoard):
 
     # Get the AMC
     amcBoard = vfatBoard.parentOH.parentAMC
+    nVFATs = vfatBoard.parentOH.nVFATs
```
bdorney, line 20: In many places you are using `vfatsPerGemVariant[args.gemType]` but here we are using the class data member.  Is there a reason for this inconsistency? e.g. are there some use cases where the class doesn't exist and so it would not be appropriate to take from this after initialization of a `HwVFAT` or `HwOptohybrid` object?  

Otherwise line 19 could probably be put in every script modified above and it might improve readability.  

I do not feel too strongly about this though.

bdorney:
Max array size still at 3072; if there is a corresponding change in `xhal` and `ctp7_modules` to make this array size smaller and an `assert` statement this could cause some errors at runtime no?

https://github.com/mexanick/vfatqc-python-scripts/blob/c8c5a67d712dbf402e6d2a0eca13adf9fcb29d0f/trimChamberV3.py#L168-L170

bdorney:
Also some function documentation and also help menu comments give info on `vfatMask` being a 24 bit number.  This should probably be changed to "at most a 24 bit number" or "either a 12 or 24 bit number depending on gemType" or similar.

Need `vfatsPerGemVariant[gemType]` in the following locations:

https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/testConnectivity.py#L172
https://github.com/mexanick/vfatqc-python-scripts/blob/c8c5a67d712dbf402e6d2a0eca13adf9fcb29d0f/monitorTemperatures.py#L86-L87
https://github.com/mexanick/vfatqc-python-scripts/blob/c8c5a67d712dbf402e6d2a0eca13adf9fcb29d0f/monitorTemperatures.py#L119-L120
https://github.com/mexanick/vfatqc-python-scripts/blob/c8c5a67d712dbf402e6d2a0eca13adf9fcb29d0f/monitorTemperatures.py#L159-L160
https://github.com/mexanick/vfatqc-python-scripts/blob/c8c5a67d712dbf402e6d2a0eca13adf9fcb29d0f/getCalInfoFromDB.py#L23-L24
https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/checkSbitMappingAndRate.py#L136
https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/checkSbitMappingAndRate.py#L241
https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/utils/confUtils.py#L355

Additionally it seems `run_scans.py` has been overlooked:

https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/run_scans.py

e.g. there should be a `parent_parser` that provides the `gemType` and `detType` arguments to all sub commands and then the functions for each sub command should transfer these arguments to the relative command.

The following functions need arguments for `gemType` and `detType`:

https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/utils/phaseUtils.py#L82

Need the `gemType` and `detType` options for the following scripts:

- https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/getCalInfoFromDB.py
- https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/monitorTemperatures.py
- https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/dacScanV3.py
- https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/confChamber.py
- https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/confAllChambers.py
- https://github.com/mexanick/vfatqc-python-scripts/blob/feature/ge21-compatibility/checkSbitMappingAndRate.py

The best way to resolve this if probably to have a `parent_parser` defined with `argparse.ArgumentParser` in `utils/qcoptions.py` and import this `parent_parser` (and any others where appropriate) to the tools that use `ArgumentParser` for their options menus.  Tools still using `optparser` will of course need a manual command line option added (or be migrated to `argparse`).

bdorney (dismissed):
still some incompatibility issues as described here; `run_scans.py` has been overlooked.  Change GE2/1 GBT config files to "generic" symlink rather than hard coded filename/path.

mexanick:
Thanks for the comments, indeed, few scripts has been overlooked. Working on them now

mexanick:
> still some incompatibility issues as described here; `run_scans.py` has been overlooked. Change GE2/1 GBT config files to "generic" symlink rather than hard coded filename/path.

I have 
```python
     if args.gemType == "ge11":
         gbtConfigPath = "{0}/OHv3c/".format(os.getenv("GBT_SETTINGS")) # Ideally this would be a DB read...
     elif args.gemType == "ge21":
         gbtConfigPath = "{0}/OHGE21/".format(os.getenv("GBT_SETTINGS"))
```

Can you clarify what do you mean under "generic" then?

utils/scanUtils.py
mexanick: I've tried to use the `vfatsPerGemVariant[args.gemType]` where `HwVFAT` object does not exist, but might have overlooked in some places

bdorney:
> > still some incompatibility issues as described here; `run_scans.py` has been overlooked. Change GE2/1 GBT config files to "generic" symlink rather than hard coded filename/path.
> 
> I have
> 
> ```python
>      if args.gemType == "ge11":
>          gbtConfigPath = "{0}/OHv3c/".format(os.getenv("GBT_SETTINGS")) # Ideally this would be a DB read...
>      elif args.gemType == "ge21":
>          gbtConfigPath = "{0}/OHGE21/".format(os.getenv("GBT_SETTINGS"))
> ```
> 
> Can you clarify what do you mean under "generic" then?

so [here](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/265#discussion_r297094218) you are referencing an exit filename; but what I'm using:

```bash
$ ll $GBT_SETTINGS/OHv3c/
total 24K
drwxr-sr--. 2 gemuser gemuser 4.0K Apr 10 15:04 20180717
lrwxrwxrwx. 1 gemuser gemuser   56 May 28 09:56 GBTX_OHv3c_GBT_1.txt -> 20180314/GBTX_OHv3c_GBT_1__2018-03-14_FINAL-REG35-62.txt
lrwxrwxrwx. 1 gemuser gemuser   56 May 28 09:56 GBTX_OHv3c_GBT_2.txt -> 20180314/GBTX_OHv3c_GBT_2__2018-03-14_FINAL-REG35-62.txt
drwxr-sr--. 2 gemuser gemuser 4.0K Jun 14 19:17 20180314
drwxr-sr--. 2 gemuser gemuser 4.0K Jun 14 19:17 20190508
lrwxrwxrwx. 1 gemuser gemuser   56 Jun 17 15:17 GBTX_OHv3c_GBT_0.txt -> 20180314/GBTX_OHv3c_GBT_0__2018-03-14_FINAL-REG35-62.txt
```

So here under the `$ENV` variable I have a set of symlinks, `GBTX_OHv3c_GBT_*.txt` that point to the exact file we need to use.

And in many cases we've had to change the file(s) that those symlinks point too; and in this way such a chance is a *system* change not a codebase change.

This is what I was trying to get at and I think you should use a similar strategy here.

jsturdy (dismissed):
cursory

testConnectivity.py
bdorney: Still outstanding

testConnectivity.py
bdorney: Still outstanding

confChamber.py
```diff
@@ -47,17 +48,19 @@
                       help="VThreshold1 DAC bump value for all VFATs", default=0)
     parser.add_argument("--zeroChan", action="store_true", dest="zeroChan",
                       help="Zero all channel registers")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
```
bdorney, line 53: Defaulting this to `None` should raise an `OHTypeException` based on these lines:

https://github.com/cms-gem-daq-project/cmsgemos/blob/48e2f269dac0947be0101d2a9f0d943752e2ce61/gempython/tools/vfat_user_functions_xhal.py#L21-L22

When you use this at line 61 I expect an error to be thrown.

dacScanV3.py
```diff
@@ -30,10 +31,12 @@
             help = "Specify output filename to store data in.")
     parser.add_argument("--series", action="store_true", dest="series",
             help = "Scan nonzero links in ohMask in series (successive RPC calls) instead of in parallel (one RPC call)")
-    parser.add_argument("--stepSize", type=int, dest="stepSize",default=1, 
+    parser.add_argument("--stepSize", type=int, dest="stepSize",default=1,
                   help="Supply a step size for the scan")
     parser.add_argument("-v","--vfatmask",type=parseInt,dest="vfatmask",default=None,
             help="VFATs to be masked in scan & analysis applications (e.g. 0xFFFFF masks all VFATs)")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
```
bdorney, line 105: [same comment](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/265/files#r300986304)

getCalInfoFromDB.py
```diff
@@ -2,25 +2,28 @@
...
     parser.add_argument("shelf",type=int,help="uTCA shelf number")
     parser.add_argument("slot",type=int,help="AMC slot in uTCA shelf")
     parser.add_argument("link",type=int,help="OH on AMC slot")
     parser.add_argument("-d","--debug",action="store_true",help="Prints additional information")
     parser.add_argument("--write2File",action="store_true",help="If Provided data will be written to appropriate calibration files")
     parser.add_argument("--write2CTP7",action="store_true",help="If Provided IREF data will be sent to the VFAT3 config files on the CTP7")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
```
bdorney: [same comment](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/265/files#r300986304)

monitorTemperatures.py
```diff
@@ -34,6 +35,8 @@
             help = "Do not print VFAT temperatures to terminal; if --filename isp provided VFAT temperatures are still read and stored in the output file")
     parser.add_argument("-t","--timeInterval", type=int, dest="timeInterval", default=60,
             help ="Time interval, in seconds, to wait in between temperature reads",metavar="timeInterval")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
```
bdorney: [same comment](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/265/files#r300986304)

run_scans.py
```diff
@@ -533,6 +552,8 @@ def ultraThreshold(args):
 
     # Option arguments shared by all commands
     parser.add_argument("-d","--debug", action="store_true",help = "Print additional debugging information")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
```
bdorney, line 708: [same comment](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/265/files#r300986304)

sbitReadOut.py
```diff
@@ -29,6 +29,10 @@
             help="Set up for using AMC13 T3 trigger inpiut")
     parser.add_argument("--vfatmask", type=parseInt, dest="vfatmask",default=0x0,
             help="VFATs to be masked, a 1 in the N^th bit signifies the N^th vfat will not be used", metavar="vfatmask")
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,
+                        help="Detector type within gemType. If gemType is 'ge11' then this should be from list {0}; if gemType is 'ge21' then this should be from list {1}; and if type is 'me0' then this should be from the list {2}".format(gemVariants['ge11'],gemVariants['ge21'],gemVariants['me0']),default=None)
```
bdorney: [same comment](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/265/files#r300986304)

sbitThreshScanParallel.py
```diff
@@ -32,28 +32,32 @@
     parser.add_argument("--scanmin", type=int, help="Minimum value of scan parameter", default=0)
     parser.add_argument("--scanmax", type=int, help="Maximum value of scan parameter", default=255)
     parser.add_argument("--stepSize", type=int, help="Supply a step size to the scan from scanmin to scanmax", default=1)
+    parser.add_argument("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_argument("--detType",type=str,
```
bdorney: [same comment](https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/265/files#r300986304)

testConnectivity.py
```diff
@@ -58,7 +58,7 @@ def getListOfBadTrigLinks(amcBoard,checkCSCTrigLink=False,debug=False,ohMask=0xf
 
     return listOfOHsWithBadTriggerLink
 
-def gbtCommIsGood(amcBoard, doReset=True, printSummary=True, ohMask=0xfff):
+def gbtCommIsGood(amcBoard, doReset=True, printSummary=True, ohMask=0xfff, gemType="ge11"):
```
bdorney, line 61: It doesn't seem like the argument `gemType` is actually used in this function.  Given that `amcBoard` is being passed as argument the type is already defined as a class member there.

testConnectivity.py
```diff
@@ -220,7 +220,7 @@ def testConnectivity(args):
     if hasattr(args, 'stepSize') is False:
         args.stepSize = 1
     if hasattr(args, 'gemType') is False:
-        args.gemType = None # default to None
+        args.gemType = "ge11" # default to ge11
```
bdorney, line 230: I think line 209 might need to be changed to have default `detType` changed to `"short"`?

testConnectivity.py
```diff
@@ -275,25 +283,35 @@ def testConnectivity(args):
...
                 "{0}/GBTX_OHv3c_GBT_2.txt".format(gbtConfigPath),
                 ]
+        elif args.gemType == "ge21":
+            gbtConfigs = [
+                "{0}/GBTX_GE21_OHv1_GBT_0.txt".format(gbtConfigPath),
+                "{0}/GBTX_GE21_OHv1_GBT_1.txt".format(gbtConfigPath),
+                ]
+        else:
+            print("me0 gemType not currently implemented, exiting.")
+            exit(os.EX_USAGE)
```
bdorney, line 312: Following some comments from @jsturdy I would change line 305 to:

```python
import sys
sys.exit(os.EX_USAGE)
```

This will raise a `SystemExit` exception which can be handled later and is necessary for our py2 to py3 migration in the end.

trimChamberV3.py
```diff
@@ -46,11 +47,14 @@
                       help="comma separated list of trim values to use in trimming, a set of scurves will be taken at each point", metavar="trimPoints")
     parser.add_option("--vfatConfig", type="string", dest="vfatConfig", default=None,
                       help="Specify file containing VFAT settings from anaUltraThreshold", metavar="vfatConfig")
-    parser.add_option("--voltageStepPulse", action="store_true",dest="voltageStepPulse", 
-                      help="Calibration Module is set to use voltage step pulsing instead of default current pulse injection", 
+    parser.add_option("--voltageStepPulse", action="store_true",dest="voltageStepPulse",
+                      help="Calibration Module is set to use voltage step pulsing instead of default current pulse injection",
                       metavar="voltageStepPulse")
+    parser.add_option("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_option("--detType",type=str,
```
bdorney, line 51: same as all the others.

ultraLatency.py
```diff
@@ -43,52 +44,54 @@
...
-                      help="Calibration Module is set to use voltage step pulsing instead of default current pulse injection", 
+    parser.add_option("--voltageStepPulse", action="store_true",dest="voltageStepPulse",
+                      help="Calibration Module is set to use voltage step pulsing instead of default current pulse injection",
                       metavar="voltageStepPulse")
     parser.add_option("--vt2", type="int", dest="vt2",
-                      help="VThreshold2 DAC value for all VFATs (v2b electronics only)", metavar="vt2", default=0) 
-
+                      help="VThreshold2 DAC value for all VFATs (v2b electronics only)", metavar="vt2", default=0)
+    parser.add_option("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_option("--detType",type=str,
```
bdorney, line 173: same

ultraScurve.py
```diff
@@ -33,20 +34,24 @@
                       help="Specify MSPL. Must be in the range 0-7 (default is 3)", metavar="MSPL")
     parser.add_option("--pulseDelay", type="int", dest = "pDel", default = 40,
                       help="Specify time of pulse before L1A in bx", metavar="pDel")
-    parser.add_option("--voltageStepPulse", action="store_true",dest="voltageStepPulse", 
-                      help="V3 electronics only. Calibration Module is set to use voltage step pulsing instead of default current pulse injection", 
+    parser.add_option("--voltageStepPulse", action="store_true",dest="voltageStepPulse",
+                      help="V3 electronics only. Calibration Module is set to use voltage step pulsing instead of default current pulse injection",
                       metavar="voltageStepPulse")
+    parser.add_option("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_option("--detType",type=str,
```
bdorney, line 122: same

ultraThreshold.py
```diff
@@ -31,18 +32,21 @@
                       help="Specify VT2 to use", metavar="vt2")
     parser.add_option("--zcc", action="store_true", dest="scanZCC",
                       help="V3 Electronics only, scan the threshold on the ZCC instead of the ARM comparator", metavar="scanZCC")
-    
+    parser.add_option("--gemType",type=str,help="String that defines the GEM variant, available from the list: {0}".format(gemVariants.keys()),default="ge11")
+    parser.add_option("--detType",type=str,
```
bdorney, line 109: same

bdorney (changes requested):
I think the default choice of `args.detType` will raise an `OHTypeException` when initializing an `HwVFAT` object.  Additionally it would have been nicer to see this added as a parent parser to `qcoptions.py` and then the parent parser used everywhere rather than a copy/paste of arguments.  Now sadly we would need to change things *everywhere* rather than one single location.

bdorney:
Needs `rebase`.

jsturdy:
bump @mexanick

mexanick:
Rebased, but needs testing, please do not merge @jsturdy

jsturdy:
If, as a result of the camp, the bifurcation between GE1/1 and GE2/1 SW versions could be resolved, I would consider that a major win

lpetre-ulb:
> If, as a result of the camp, the bifurcation between GE1/1 and GE2/1 SW versions could be resolved, I would consider that a major win

During my GE2/1 GLIB tests, this PR ended-up working very well. However, I reverted to the main branch (`rpc-playground`), because of a usability flaw. Indeed, the `--gemType` and `--detType` must be specified for every command.

* `--gemType` can be retrieved from the register `GEM_AMC.GEM_SYSTEM.RELEASE.GEM_STATION` starting from GEM_AMC release 3.9.1. For newer release, the GEM station should always be checked against the firmware register.
* `--detType` should be retrieved from the `config_chamber` dictionaries by default. I did not find a way to test both long and short GEBs (GE1/1) with `testConnectivity.py`. This is main reason why we are using the main branch and do not test this PR at ULB.

This PR also powers the FIT setup. @sbutalla, are you encountering any other issues?

jsturdy:
> Merged with current tip of central   `0f4af43` 

should have been a `rebase` rather than a `merge`

jsturdy:
> During my GE2/1 GLIB tests, this PR ended-up working very well. However, I reverted to the main branch (`rpc-playground`), because of a usability flaw. Indeed, the `--gemType` and `--detType` must be specified for every command.

As a stop-gap solution that this was foreseen to be, this is acceptable (provided that the requirement that they be provided exists **only** for `GE2/1` operations)

> 
>     * `--gemType` can be retrieved from the register `GEM_AMC.GEM_SYSTEM.RELEASE.GEM_STATION` starting from GEM_AMC release 3.9.1. For newer release, the GEM station should always be checked against the firmware register.
> 
>     * `--detType` should be retrieved from the `config_chamber` dictionaries by default. I did not find a way to test both long and short GEBs (GE1/1) with `testConnectivity.py`. This is main reason why we are using the main branch and do not test this PR at ULB.

Very little effort should be spent trying to reconcile `detType` and `chamber_config`, and any real thought spent on thinking how best to do this long term through the configuration DB (or text proxy of such, @lmoureaux )

lmoureaux:
> Very little effort should be spent trying to reconcile `detType` and `chamber_config`, and any real thought spent on thinking how best to do this long term through the configuration DB (or text proxy of such, @lmoureaux )

Currently, `gemonlinedb` does not know about the composition of the chambers. Pulling this info from the DB would be easy, and it could also be added to the XML files. This is however a second-order priority with respect to having a working GE1/1 setup.

Generally speaking, I think that information about the hardware layout should either be discovered automatically or come from configuration files. Selecting which link/chamber/... to use is a valid use-case for a command line argument, while telling the system which kind of hardware it is dealing with is not. Just like your OS doesn't ask whether your disk is HDD or SSD every time you read or write data :stuck_out_tongue:

jsturdy:
> Generally speaking, I think that information about the hardware layout should either be discovered automatically or come from configuration files. Selecting which link/chamber/... to use is a valid use-case for a command line argument, while telling the system which kind of hardware it is dealing with is not. Just like your OS doesn't ask whether your disk is HDD or SSD every time you read or write data stuck_out_tongue

Generally I agree with this, and prefer that where functionality depends on these parameters, they should come from the HW itself, while for random tracking metadata, it *could* come from the configuration (this avoids doing additional register requests for **every single** operation, even when the behaviour doesn't change between HW variants)

mexanick:
Generally speaking, things are fixed now, apart from the configuration/detector types. But that should be addressed in a different issue I believe...

lpetre-ulb:
> Generally speaking, things are fixed now, apart from the configuration/detector types.

I believe this PR should be the latest bringing new features to be merged in the legacy branch. The work is already done and the features are important for the GE2/1 US test stands. The PR has been tested at CERN and FIT and seem to work fine.

Fixing the configuration/detector type concerns is out of the scope of this PR and probably out of the scope of the `legacy` branch.

What do you think @mexanick, @jsturdy?

jsturdy:
just rebase it on top of either the current target branch, or the new legacy branch (and change the target)

mexanick:
ok, rebased and retargeted. I believe things *should* work OK, not 100% sure in trimming (although if doesn't work, changes should be very simple and straightforward). Plan to test today with GE2/1 installation at the QC7.

jsturdy (approved):
Standard legacy cursory

jsturdy:
> ok, rebased and retargeted. I believe things _should_ work OK, not 100% sure in trimming (although if doesn't work, changes should be very simple and straightforward). Plan to test today with GE2/1 installation at the QC7.

Did this get tested? If so (and it passed), please merge

mexanick:
In process of testing, will be completed today. In principle, everything is expected to work out of the box except perhaps iterative trimming, but this will be tested today.

confChamber.py
lpetre-ulb: Isn't this comment still relevant? For example, the behavior in `checkSbitMappingAndRate.py` is different: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/a9637d6c92d4336eeebcd4ac0aa18cdd77606465/checkSbitMappingAndRate.py#L109

monitorTemperatures.py
```diff
@@ -148,8 +151,8 @@
 
     from time import sleep, time
     from ctypes import *
-    adcDataIntRefMultiLinks = (c_uint32 * (24 * 12))()
-    adcDataExtRefMultiLinks = (c_uint32 * (24 * 12))()
+    adcDataIntRefMultiLinks = (c_uint32 * (vfatBoard.parentOH.nVFATs * 12))()
+    adcDataExtRefMultiLinks = (c_uint32 * (vfatBoard.parentOH.nVFATs * 12))()
```
lpetre-ulb, line 155: Do the legacy RPC modules and xHAL interface really support a variable number of VFAT? It is not always the case.

testConnectivity.py
lpetre-ulb: Doesn't it makes sense to review this argument as suggested by Brian?

lpetre-ulb (approved):
Just a few comments (+ some left by Brian). Nothing outstanding which would block this PR.

mexanick:
The majority of tools are working just fine. Some tools which haven't been used by anyone for ages may exhibit weird behavior. I'd merge it and fix any other issues as soon as they arise

lpetre-ulb:
> The majority of tools are working just fine. Some tools which haven't been used by anyone for ages may exhibit weird behavior. I'd merge it and fix any other issues as soon as they arise

Agreed! The few remaining comments are not outstanding. As soon as the merge is done, a new release should be produced and published.

[I would merge it myself, but I'm not sure what's the merge policy. It also looks like it is different between repos.]

jsturdy:
`2.8.0`: there's no way that these changes could fit into a patch release by any stretch of the imagination
or 
`2.7.12`: there's also likely never been any strict adherence to semver since someone made an initial ninja tag, thereby forcing all subsequent tags to follow suit

lpetre-ulb:
Since the legacy branch is named `legacy-2.7`, I would go for `2.7.12`. As you wrote, semver has never been strictly followed...

If you were to use the more semantically correct version `2.8.0`, I would rename the legacy branch at the same time. This is also the lastest new feature that is envisioned for legacy software.

PR #268, opened 2019-07-04
Ignore uHAL exceptions regarding the ADC monitoring block (partially fixes #266)

### LINKED ISSUE #266: [Feature Request] Update Legacy Branch to Not Read/Write to SCA.ADC_Monitoring Block
## Brief summary of issue

The `SCA.ADC_MONITORING` block was removed from GEM_AMC in [v3.8.4](https://github.com/evka85/GEM_AMC/releases/tag/v3.8.4).

However not all test stands have migrated to this version yet due to issues with OH FW.  We need to have some protection regarding writing/reading from these nodes

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

In locations below we could either:

1. Do a read of the AMC FW and if below 3.8.4 then the read/write this block as normal; if above 3.8.4 do nothing, or
2. Put all read/writes to this block in a `try...except` block and try to catch either `uhal._core.exception`.

If a node name is not found in the xhal/reg_utils equivalent command no exception is raised so no handling is "needed." Given this branch is legacy anyway I doubt something further is needed.

## Current Behavior

The `SCA.ADC_MONITORING` block is presently used in two places:

```bash
% grep -R "SCA.ADC_MONITORING" *.py
monitorTemperatures.py:    origSCAMonOffVal = amcBoard.readRegister("GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF")
testConnectivity.py:    writeRegister(amc,"GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF",0xffffffff)
```

## Context (for feature requests)

Need software to be able to use either FW regardless of release. 

## Your Environment

* Version used: dae6fb9a1d65f0d7081dc040832faf1c5f77123e
* Shell used:

lpetre-ulb:
## Description

Any uHAL exception occurring while trying to disable the ADC monitoring block are now ignored with an explicit warning message.

This is the simplest option and has been chosen since the target branch is legacy. Moreover issues to connect to the CTP7 IPBus daemon are already caught earlier in the script. Finally all code regarding the ADC monitoring block will be removed in the future, once the firmwares will be stable. 

### Types of changes

## Motivation and Context

See issue #266.

## How Has This Been Tested?

This patch was extensively tested during GEB&OH QC at ULB with the CTP7 firmware version 3.8.5.

## Checklist:

### DIFF
--- a/testConnectivity.py
+++ b/testConnectivity.py
@@ -97,7 +97,12 @@ def scaCommIsGood(amc, maxIter=5, ohMask=0xfff, nOHs=12):
     from reg_utils.reg_interface.common.jtag import initJtagRegAddrs
     initJtagRegAddrs()
 
-    writeRegister(amc,"GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF",0xffffffff)
+    try:
+        writeRegister(amc,"GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF",0xffffffff)
+    except uhal._core.exception:
+        printYellow("An exception has been caught while attempting to disable the ADC monitoring.")
+        printYellow("If you use a CTP7 with a firmware version higher than 3.8.3 you can safely ignore this warning.")
+        pass
 
     for trial in range(0,maxIter):
         sca_reset(ohMask)


jsturdy:
Do you plan to also patch `monitorTemperatures.py`?

lpetre-ulb:
Not immediately since it requires the SCA temperature RPC calls (or writing dummy data to the some `TBranch`). I guess it would be patched in the series of patches that @mexanick is working on.

jsturdy:
But as I understand it, this change will *only* prevent the error messages, it does not implement the different mechanism for value retrieval

lpetre-ulb:
> But as I understand it, this change will _only_ prevent the error messages, it does not implement the different mechanism for value retrieval

Indeed it only prevents the non-existing register in `uHAL` to crash the `testConnectivity.py` script; there is no value to retrieve though, only one write to perform.

For the `monitorTemperatures.py` script, the same method _could_ be applied. However data need to be read and stored. Sure one could catch all exceptions are write dummy values to the ROOT Tree but waiting for @mexanick patch seems so much more useful.

bdorney:
Right now the SCA values read by `monitorTemperatures.py` are garbage anyway due to issues in the implementation of the FW.

To address this we should either backport the functionality from `develop` of `ctp7_modules` or drop this entirely.  Given that:

https://github.com/cms-gem-daq-project/ctp7_modules/pull/137

This is in review we could foresee the requirement that `monitorTemperatures.py` also be updated here.  This however would require an additional PR to legacy branches of `cmsgemos` and `xhal`.

Is this acceptable @jsturdy ?

bdorney (approved):

jsturdy:
> Right now the SCA values read by `monitorTemperatures.py` are garbage anyway due to issues in the implementation of the FW.
> 
> To address this we should either backport the functionality from `develop` of `ctp7_modules` or drop this entirely. Given that:
> 
> [cms-gem-daq-project/ctp7_modules#137](https://github.com/cms-gem-daq-project/ctp7_modules/pull/137)
> 
> This is in review we could foresee the requirement that `monitorTemperatures.py` also be updated here. This however would require an additional PR to legacy branches of `cmsgemos` and `xhal`.
> 
> Is this acceptable @jsturdy ?

Probably, provided that my understanding that changing the two lines that currently reference this register in `monitorTemperatures` *can't* be simply protected as they are in `testConnectivity` with no change in functionality, which was the point of my comment

jsturdy (approved):

PR #271, opened 2019-07-12
New Feature: Additional Configure Options for `confAllChambers.py`

bdorney:
## Description

In some cases we want channel registers applied (e.g. to determine `CFG_THR_ARM_DAC` setting without hot channels throwing off the SBIT threshold measurements).  In other cases we want only `CFG_THR_ARM_DAC` applied and no channel registers applied.

### Types of changes

Breaking because the `--config` option has been replaced by several sets of config options:

```
% confAllChambers.py -h
usage: confAllChambers.py [-h] [--armDAC ARMDAC] [--armDACBump ARMDACBUMP]
                          [-d] [--run] [--series] [--shelf SHELF]
                          [--applyMasks] [--vfatConfig]
                          [--chConfig | --zeroChan]

Tool for configuring all front-end electronics

optional arguments:
  -h, --help            show this help message and exit
  --armDAC ARMDAC       CFG_THR_ARM_DAC value to write to all VFATs
  --armDACBump ARMDACBUMP
                        CFG_THR_ARM_DAC value for all VFATs
  -d, --debug           prints additional debugging information
  --run                 Set VFATs to run mode
  --series              Run tests in series (default is false)
  --shelf SHELF         uTCA shelf number

Configuration Group:
  Options for configuring channel registers and CFG_THR_ARM_DAC

  --applyMasks          If paired with --chConfig channel masks defined in
                        chConfig text file will be applied; otherwise no
                        effect
  --vfatConfig          Set only CFG_THR_ARM_DAC registers from symlinks found
                        under $DATA_PATH/configs
  --chConfig            Set only channel registers from symlinks found under
                        $DATA_PATH/configs
  --zeroChan            Zeros all channel registers
```

## Motivation and Context

Need additional configuration options for working with QC8.

## How Has This Been Tested?

On the QC8 test stand.

## Checklist:

### DIFF
--- a/confAllChambers.py
+++ b/confAllChambers.py
@@ -30,23 +30,23 @@ def launchArgs(shelf,slot,link,run,armDAC,armDACBump,configType,cName,debug=Fals
             )
 
     from gempython.utils.gemlogger import printYellow
-    if ((configType & 0x1) == 0x1):         # Set vfatConfig
+    if (configType & 0x1):          # Set vfatConfig
         vfatConfig = "{0}/configs/vfatConfig_{1}.txt".format(dataPath,cName)
 
         if os.path.isfile(vfatConfig):
             args.vfatConfig = vfatConfig
         else:
             printYellow("No vfat configuration exists for {0}".format(cName))
-    if (((configType >> 1) & 0x1) == 0x1):  # Set chConfig and potentially channel masks
+    if ( (configType & 0x2) > 0):   # Set chConfig and potentially channel masks
         chConfig = "{0}/configs/chConfig_{1}.txt".format(dataPath,cName)
 
         if os.path.isfile(chConfig):
             args.chConfig = chConfig
-            if (((configType >> 2) & 0x1) == 0x1):
+            if ( (configType & 0x4) > 0):
                 args.applyMasks = True
         else:
             printYellow("No channel configuration exists for {0}".format(cName))
-    if (((configType >> 3) & 0x1) == 0x1):  # Zero all channel registers
+    if ( (configType & 0x8) > 0):   # Zero all channel registers
         args.chConfig = None
         args.applyMasks = False
         args.zeroChan = True
@@ -95,13 +95,13 @@ if __name__ == '__main__':
     # [3] -> zero channels
     configType=0x0
     if args.vfatConfig:
-        configType += 0x1
+        configType |= 0x1
     if args.chConfig:
-        configType += (0x1 << 1)
+        configType |= 0x2
         if args.applyMasks:
-            configType += (0x1 << 2)
+            configType |= 0x4
     if args.zeroChan:
-        configType += (0x1 << 3)
+        configType |= 0x8
         pass
 
     # consider only the shelf of interest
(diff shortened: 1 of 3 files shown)


bdorney:
@jsturdy now following our discussion from SW Dev Meeting and on mattermost; tested and ready for review.

confAllChambers.py
```diff
@@ -28,22 +29,27 @@ def launchArgs(shelf,slot,link,run,armDAC,armDACBump,config,cName,debug=False):
...
-        else:
-            print("No channel configuration exists for {0}".format(cName))
-        
-        # VFAT Config
         if os.path.isfile(vfatConfig):
             args.vfatConfig = vfatConfig
         else:
-            print("No vfat configuration exists for {0}".format(cName))
+            printYellow("No vfat configuration exists for {0}".format(cName))
+    if (((configType >> 1) & 0x1) == 0x1):  # Set chConfig and potentially channel masks
```
jsturdy: No need for bitshifting here and equality, just check
```diff
-if (configType&0x10 > 0)
+if (configType&0x2 > 0)
```
~~Similarly  in L45 (`0x100`) , L49 (`0x1000`), and L33 (`> 0` consistency)~~
of course, what I actually meant was:
Similarly  in L45 (`0x4`) , L49 (`0x8`), and L33 (`> 0` consistency)
(edited)

confAllChambers.py
```diff
@@ -66,16 +72,38 @@ def launchArgs(shelf,slot,link,run,armDAC,armDACBump,config,cName,debug=False):
...
+    # determine configType, 4-bit number with bit meaning as:
+    # [0] -> apply vfatConfig
+    # [1] -> apply chConfig
+    # [2] -> apply channel masks
+    # [3] -> zero channels
+    configType=0x0
+    if args.vfatConfig:
+        configType += 0x1
+    if args.chConfig:
+        configType += (0x1 << 1)
```
jsturdy: Similarly, here you can use:
```diff
-configType |= 0x10
+configType |= 0x2
```

You could define a set of enums that define the meaning of each bit, so that both here and above, the intent is clear from the code (but this is not strictly necessary)
(edited)

jsturdy (changes requested):
A few simplifications

confAllChambers.py
bdorney: Hah, when I originally read this I read it as binary...only now catching the hex issue. But yeah understood.

confAllChambers.py
bdorney: Updated using bitwise operators.

> You could define a set of enums that define the meaning of each bit, so that both here and above, the intent is clear from the code (but this is not strictly necessary)

If this is not necessary I would prefer to wait on this for either migration to python3 when the `enum` type will be available and/or when we move to `boost`ing things from `cmsgemos` and have this in a dedicated header file in a single location.

bdorney:
@jsturdy requested changes addressed.

jsturdy (approved):
:+1:

mexanick (approved):

PR #272, opened 2019-08-22
New Feature: Iterative Trimming Algorithm

bdorney:
## Description

Provided a new tool `iterativeTrim.py` based on algorithm presented [here](https://indico.cern.ch/event/838248/contributions/3515801/attachments/1887448/3112771/VFAT3b_trim.pdf).

### Types of changes

## Motivation and Context

Needed a faster and more reliable trimming algorithm which can enable us to push the thresholds to lower values.

## How Has This Been Tested?

```bash
% iterativeTrim.py -h
usage: iterativeTrim.py [-h] [--calFileCAL CALFILECAL] [--calSF CALSF]
                        [--chMin CHMIN] [--chMax CHMAX] [-c]
                        [--dirPath DIRPATH] [-d] [-l LATENCY] [-m MAXITER]
                        [-n NEVTS] [-p PULSESTRETCH] [-v VFATMASK] [-z]
                        shelf slot link

Arguments to supply to iterativeTrim.py

positional arguments:
  shelf                 uTCA crate shelf number
  slot                  AMC slot number in the uTCA crate
  link                  OH number on the AMC slot of interest

optional arguments:
  -h, --help            show this help message and exit
  --calFileCAL CALFILECAL
                        File specifying calibration constants for CFG_CAL_DAC;
                        if not provided a DB query will be performed at
                        runtime to extract this information based on VFAT
                        ChipIDs
  --calSF CALSF         CFG_CAL_FS value to be used; no effect if
                        --currentPulse is not supplied
  --chMin CHMIN         Specify minimum channel number to scan
  --chMax CHMAX         Specify maximum channel number to scan
  -c, --currentPulse    Operate calibration mode in current pulse injection
                        rather then voltage step pulse injection
  --dirPath DIRPATH     Specify the path where the scan data should be stored
  -d, --debug           Prints additional debugging information
  -l LATENCY, --latency LATENCY
                        CFG_LATENCY value to be used
  -m MAXITER, --maxIter MAXITER
                        Maximum number of iterations to perform (e.g. number
                        of scurves to take)
  -n NEVTS, --nevts NEVTS
                        Number of events for each scan position
  -p PULSESTRETCH, --pulseStretch PULSESTRETCH
                        CFG_PULSE_STRETCH value to be used
  -v VFATMASK, --vfatmask VFATMASK
                        Specifies which VFATs, if any, should be masked. Here
                        this is a 24 bit number, where a 1 in the N^th bit
                        means ignore the N^th VFAT.
  -z, --zeroChan        Zero all channel registers before beginning iterative
                        trim procedure
```

Usage example:

```
iterativeTrim.py 1 2 10 -m 3 -z
```

Here this will launch the iterative trimming procedure for geographic address `(shelf,slot,link) = (1,2,10)` this will make three iterations and this will start by zero'ing all channel registers.

If you'd like to add *more* iterations or resume a previous iterative trimming run from the last good iterative you can use the `--dirPath` argument, e.g.

```
iterativeTrim.py 1 2 10 -m 6 -z --dirPath=/path/to/previous/location
``` 

Here if the previous call was with `-m 3` then three additional iterations will be added; if the previous call failed this will look for the *last* good iteration and start from there.

Note the `SCurveData_iter3.root` file is the scurve taken in the third iteration but the channel configuration registers will have been used from the second iteration.

So to see the effect of the iterative process you need to take another scurve configured at the chConfig file produced from your *final* iteration.

You can see the effectiveness here:

**Untrimmed**
Screen Shot 2019-08-22 at 4 45 55 PM
h2ScurveMeanDist_All

**Trimmed**, this is using the settings derived from the third scurve.
Screen Shot 2019-08-22 at 4 46 06 PM
h2ScurveMeanDist_All

Here you can see the spread in scurve mean (threshold) has reduced after using the settings after the third iteration:
Screen Shot 2019-08-22 at 4 45 46 PM

Here you can see the effect indeed enables the `CFG_THR_ARM_DAC` register to be lowered and that the pedestal is shifted to lower values:

(Using the `chConfig` settings from the third iteration)
Screen Shot 2019-08-22 at 4 47 16 PM

After the third iteration I did not see a significant improvement in the results so one could simply stop after three iterations.  I haven't done this on enough detectors to know what a good automatic exit condition is so for now I would simply recommend stopping at three iterations.

## Checklist:

- [x] I have updated the documentation accordingly.


jsturdy:
Do you have a (qualitative or quantitative) comparison to the results from the previous algorithm?
Do you have (can you ask someone to obtain) a set of measurements taking these trim settings an moving the `THR_ARM_DAC` working point? (and then this also as a comparison with the old trim algorithm, which we fully expect to degrade)

bdorney:
> Do you have a (qualitative or quantitative) comparison to the results from the previous algorithm?

I don't have results really handy but from what we've done on QC8 I've seen that since `trimChamberV3.py` is normalizing all channels to some charge Q, rather than attempting to minimize the spread, that corresponds to a `CFG_THR_ARM_DAC` value once you move significantly away (read "outside about half the range of the `TRIM_ARM_AMPLITUDE` DAC) from Q the set of {trimDAC} values is no longer valid and just sets the channels to some random value.  We've seen this on QC8 that a `chConfig` from `trimChamberV3.py` is only valid near the `{CFG_THR_ARM_DAC}` set it was derived at. This means that the pedestal is not actually effected by trimming and there's no decrease in the noise as a result (since you're only normalized at Q+epsilon).  While when minimizing the spread of `{trimDAC}` this set of values is independent of `CFG_THR_ARM_DAC` (since indeed it's a residual) and this is what was previously shown by the VFAT3 team.  So this shift in pedestal to lower values occurs.

TL;DR: we do not see the reduction in the pedestal as observed in the sbit threshold measurement as shown above when trimming with `trimChamberV3.py`; additionally it takes significantly longer since the calibration of `CFG_THR_ARM_DAC` is a pre-requisition to trimming.

> Do you have (can you ask someone to obtain) a set of measurements taking these trim settings an moving the `THR_ARM_DAC` working point? (and then this also as a comparison with the old trim algorithm, which we fully expect to degrade)

I will see what's possible with the QC8 team; but I don't think either of these should be "blocking" or "required" for this PR.

bdorney:
> Do you have (can you ask someone to obtain) a set of measurements taking these trim settings an moving the THR_ARM_DAC working point? (and then this also as a comparison with the old trim algorithm, which we fully expect to degrade)

A good exercise for @bregnery would be to follow up today's study on `OH2` of `gem.shelf01.amc06` of the QC8 stand with the request above though.

jsturdy (approved):
cursory

PR #277, opened 2019-09-18
[New Feature] Variable waitTime for parallel sbitrate scans

AndrewLevin:
This is one of four pull requests to four repositories that make the data collection time for parallel sbitrate scans configurable.

## Description
Adds a `waitTime` argument in `run_scans.py` and `sbitThreshScanParallel.py`

### Types of changes

## Motivation and Context
This was requested in https://github.com/cms-gem-daq-project/ctp7_modules/issues/154

## How Has This Been Tested?
Yes, I performed an sbitThresh scan with the waitTime set to 1000 and the waitTime set to 10000, and found the rates were very close while the second scan took around 10 times longer.

## Checklist:

### DIFF
--- a/run_scans.py
+++ b/run_scans.py
@@ -662,7 +662,7 @@ if __name__ == '__main__':
     parser_sbitThresh.add_argument("--scanmin",type=int,default=0,help="Minimum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--scanmax",type=int,default=255,help="Maximum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--stepSize",type=int,default=1,help="Step size to use when scanning CFG_THR_ARM_DAC")
-    parser_sbitThresh.add_argument("--waitTime",type=int,default=1000,help="Length of the time window within which the rate is measured, in milliseconds")
+    parser_sbitThresh.add_argument("--waitTime",type=int,default=1000,help="Length of the time window within which the rate is measured, in seconds")
     parser_sbitThresh.set_defaults(func=sbitThreshScan)
 
 
--- a/run_scans.py
+++ b/run_scans.py
@@ -662,7 +662,7 @@ if __name__ == '__main__':
     parser_sbitThresh.add_argument("--scanmin",type=int,default=0,help="Minimum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--scanmax",type=int,default=255,help="Maximum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--stepSize",type=int,default=1,help="Step size to use when scanning CFG_THR_ARM_DAC")
-    parser_sbitThresh.add_argument("--waitTime",type=int,default=1000,help="Length of the time window within which the rate is measured, in seconds")
+    parser_sbitThresh.add_argument("--waitTime",type=int,default=1,help="Length of the time window within which the rate is measured, in seconds")
     parser_sbitThresh.set_defaults(func=sbitThreshScan)
 
 
--- a/run_scans.py
+++ b/run_scans.py
@@ -230,6 +230,7 @@ def sbitThreshScan(args):
             "--scanmax={}".format(args.scanmax),
             "--scanmin={}".format(args.scanmin),
             "--stepSize={}".format(args.stepSize),
+            "--waitTime={}".format(args.waitTime),        
             str(args.shelf),
             str(args.slot),
             "0x{:x}".format(args.ohMask)
@@ -661,6 +662,7 @@ if __name__ == '__main__':
     parser_sbitThresh.add_argument("--scanmin",type=int,default=0,help="Minimum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--scanmax",type=int,default=255,help="Maximum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--stepSize",type=int,default=1,help="Step size to use when scanning CFG_THR_ARM_DAC")
+    parser_sbitThresh.add_argument("--waitTime",type=int,default=1000,help="Length of the time window within which the rate is measured, in milliseconds")
     parser_sbitThresh.set_defaults(func=sbitThreshScan)
 
 
--- a/sbitThreshScanParallel.py
+++ b/sbitThreshScanParallel.py
@@ -32,6 +32,7 @@ if __name__ == '__main__':
     parser.add_argument("--scanmin", type=int, help="Minimum value of scan parameter", default=0)
     parser.add_argument("--scanmax", type=int, help="Maximum value of scan parameter", default=255)
     parser.add_argument("--stepSize", type=int, help="Supply a step size to the scan from scanmin to scanmax", default=1)
+    parser.add_argument("--waitTime", type=int, help="Length of the time window within which the rate is measured, in milliseconds",default=1000)
     args = parser.parse_args()
     
     from gempython.utils.gemlogger import printRed, printYellow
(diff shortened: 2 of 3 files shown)


run_scans.py
```diff
@@ -661,6 +662,7 @@ def ultraThreshold(args):
     parser_sbitThresh.add_argument("--scanmin",type=int,default=0,help="Minimum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--scanmax",type=int,default=255,help="Maximum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--stepSize",type=int,default=1,help="Step size to use when scanning CFG_THR_ARM_DAC")
+    parser_sbitThresh.add_argument("--waitTime",type=int,default=1000,help="Length of the time window within which the rate is measured, in milliseconds")
```
jsturdy: Unless there is inertia elsewhere in the existing codebase for milleseconds, I would discretize this in seconds

AndrewLevin: Do we want to allow fractional values of seconds? I believe we are not sending any `float`s in the rpc calls and it is not clear to me whether it is possible or a good idea. But we could make the argument a `float` in seconds in this pull request and then convert it to an `int` in milliseconds for the `cmsgemos` function.

jsturdy: I suppose we could... but I think we're not interested in going below 1s.
Where it could be interesting would be if the scan itself was much more granular and could take multiple measurements to get proper uncertainties, but I think this is also not the most important thing.
What we should get in the immediate term is the ability to use the FW counters for some programmable time and determine the rate in Hz from the counter and the integration time (this last bit may only make sense to do in the analysis, since everything up to that point is just an integer type, right?)

In the CTP7 FW, the rate calculator is hard-coded to rearm after 1s, though the counter itself could be read whenever and just rolls over.
In the OH FW, the counter integrates for a programmable time and then resets

AndrewLevin: > I suppose we could... but I think we're not interested in going below 1s.
> Where it could be interesting would be if the scan itself was much more granular and could take multiple measurements to get proper uncertainties, but I think this is also not the most important thing.
> What we should get in the immediate term is the ability to use the FW counters for some programmable time and determine the rate in Hz from the counter and the integration time (this last bit may only make sense to do in the analysis, since everything up to that point is just an integer type, right?)

So, you are proposing that instead of storing a `rate` in the output tree, which is what we are doing now, we store both the `count` and the `waitTime`?

> 
> In the CTP7 FW, the rate calculator is hard-coded to rearm after 1s, though the counter itself could be read whenever and just rolls over.
> In the OH FW, the counter integrates for a programmable time and then resets

OK, I see now that we are measuring an all-VFAT rate and then 24 per-VFAT rates. It seems that the register used in the all-VFAT rate measurement ("GEM_AMC.TRIGGER.OH0.TRIGGER_RATE") is on the CTP7 while the register used in per-VFAT rate measurement ("GEM_AMC.OH.OH0.FPGA.TRIG.CNT.VFAT0_SBITS") is on the OH.

Is the register "GEM_AMC.TRIGGER.OH0.TRIGGER_RATE" actually a counter or rate?

jsturdy: > > I suppose we could... but I think we're not interested in going below 1s.
> > Where it could be interesting would be if the scan itself was much more granular and could take multiple measurements to get proper uncertainties, but I think this is also not the most important thing.
> > What we should get in the immediate term is the ability to use the FW counters for some programmable time and determine the rate in Hz from the counter and the integration time (this last bit may only make sense to do in the analysis, since everything up to that point is just an integer type, right?)
> 
> So, you are proposing that instead of storing a `rate` in the output tree, which is what we are doing now, we store both the `count` and the `waitTime`?
> 

Hmm, no, I'm proposing that we store the same thing, though I was thinking that we were storing the count as the rate, since it was integrated over 1s.
If the count was previously being divided by some wait time (of 1s) and this value being stored, then we should continue, except with the now configurable wait time.

> > In the CTP7 FW, the rate calculator is hard-coded to rearm after 1s, though the counter itself could be read whenever and just rolls over.
> > In the OH FW, the counter integrates for a programmable time and then resets
> 
> OK, I see now that we are measuring an all-VFAT rate and then 24 per-VFAT rates. It seems that the register used in the all-VFAT rate measurement ("GEM_AMC.TRIGGER.OH0.TRIGGER_RATE") is on the CTP7 while the register used in per-VFAT rate measurement ("GEM_AMC.OH.OH0.FPGA.TRIG.CNT.VFAT0_SBITS") is on the OH.
> 
> Is the register "GEM_AMC.TRIGGER.OH0.TRIGGER_RATE" actually a counter or rate?

This register is truly a rate, and it has an integration time of 1s, not programmable, though this is not as important, one could read this rate every 1+ second to account for delays, and then take the average to get the time averaged rate, but again, this particular value is less important than the per-VFAT value.
Alternatively, we can do the same thing as is done for each VFAT, since there is a CTP7 `OHX_TRIGGER_CNT` register that behaves (mostly) the same as the per-VFAT counters in the OH.
In the interest of the minimal change now, let's only focus on the per-VFAT rate in this set of PRs

AndrewLevin: ok, so the `waitTime` will be applied only to the per-VFAT counters

AndrewLevin: The count was previously not being divided by anything, which is equivalent to dividing by 1 s. Now, I think what we want to do is divide by the `waitTime`, which I have done here:

https://github.com/AndrewLevin/ctp7_modules/blob/sbitrate_waittime/src/calibration_routines.cpp#L711

AndrewLevin: Which was probably not a good idea since the right-hand side is a float while the left-hand side is an integer. Maybe it is a better idea to return the count, and then perform the division in vfatqc-python-scripts.

jsturdy: > Which was probably not a good idea since the right-hand side is a float while the left-hand side is an integer. Maybe it is a better idea to return the count, and then perform the division in vfatqc-python-scripts.

Yeah, this was what I was thinking was the case.   Returning the count, then let the thing that fills the tree take care of converting to the appropriate rate.

AndrewLevin:
I have updated this pull request and the `ctp7_modules` pull request so that the division to calculate the rate is performed here now.

run_scans.py
AndrewLevin: I have changed all of the 4 pull requests to use seconds instead of milliseconds now.

run_scans.py
```diff
@@ -661,6 +662,7 @@ def ultraThreshold(args):
     parser_sbitThresh.add_argument("--scanmin",type=int,default=0,help="Minimum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--scanmax",type=int,default=255,help="Maximum CFG_THR_ARM_DAC")
     parser_sbitThresh.add_argument("--stepSize",type=int,default=1,help="Step size to use when scanning CFG_THR_ARM_DAC")
+    parser_sbitThresh.add_argument("--waitTime",type=int,default=1000,help="Length of the time window within which the rate is measured, in seconds")
```
jsturdy: Default is 1000s?

sbitThreshScanParallel.py
```diff
@@ -32,6 +32,7 @@
     parser.add_argument("--scanmin", type=int, help="Minimum value of scan parameter", default=0)
     parser.add_argument("--scanmax", type=int, help="Maximum value of scan parameter", default=255)
     parser.add_argument("--stepSize", type=int, help="Supply a step size to the scan from scanmin to scanmax", default=1)
+    parser.add_argument("--waitTime", type=int, help="Length of the time window within which the rate is measured, in seconds",default=1000)
```
jsturdy: Also here

run_scans.py
AndrewLevin: fixed

sbitThreshScanParallel.py
AndrewLevin: fixed

jsturdy (approved):
:+1:

PR #282, opened 2019-10-15
[Bug Fix]: vfatConfig and armDac arguments were missing in iterative trimmer

AndrewLevin:
This pull requests adds `vfatConfig` and `armDAC` arguments to `iterativeTrim.py` and handles them properly.

## Description
The `armDAC` argument takes precedence over the `vfatConfig` argument. `run_scans.py` is automatically checking for vfat config files in a standard location with a standard name when `armDAC` is not provided.

### Types of changes

## Motivation and Context
When installing a software update on the QC7/8 machines, it was reported that `run_scans.py` was passing `--vfatConfig` to `iterativeTrim.py`, which was causing a crash.

## How Has This Been Tested?
Yes, I ran iterative trim runs with both the `--vfatConfig` and the `--armDAC` arguments on the coffin.

## Checklist:

### DIFF
--- a/iterativeTrim.py
+++ b/iterativeTrim.py
@@ -199,7 +199,7 @@ if __name__ == '__main__':
 
     if args.vfatConfig is not None:
         try:
-            print 'Configuring VFAT Registers based on %s'%args.vfatConfig
+            print('Configuring VFAT Registers based on {0}'.format(args.vfatConfig))
             vfatTree = r.TTree('vfatTree','Tree holding VFAT Configuration Parameters')
             vfatTree.ReadFile(args.vfatConfig)
             
@@ -209,11 +209,11 @@ if __name__ == '__main__':
                     continue
                     
                 # Write CFG_THR_ARM_DAC
-                print('Set link %d VFAT%d CFG_THR_ARM_DAC to %i'%(args.link,event.vfatN,event.vt1))
+                print('Set link {0} VFAT{1} CFG_THR_ARM_DAC to {2}'.format(args.link,event.vfatN,event.vt1))
                 vfatBoard.setVFATThreshold(chip=int(event.vfatN), vt1=int(event.vt1))
         except IOError as e:
-            print '%s does not seem to exist or is not readable'%options.filename
-            print e
+            print('{0} does not seem to exist or is not readable'.format(options.filename))
+            print(e)
     else:
         vfatBoard.setVFATThresholdAll(mask=args.vfatmask, vt1=args.armDAC)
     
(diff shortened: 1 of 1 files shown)


iterativeTrim.py
```diff
@@ -195,6 +197,26 @@ def writeChConfig(chConfig,vfat,vfatID,trimDAC,trimPol,mask,maskReason):
...
+            for event in vfatTree :
+                # Skip masked vfats
+                if (args.vfatmask >> int(event.vfatN)) & 0x1:
+                    continue
+                    
+                # Write CFG_THR_ARM_DAC
+                print('Set link %d VFAT%d CFG_THR_ARM_DAC to %i'%(args.link,event.vfatN,event.vt1))
+                vfatBoard.setVFATThreshold(chip=int(event.vfatN), vt1=int(event.vt1))
+        except IOError as e:
+            print '%s does not seem to exist or is not readable'%options.filename
```
jsturdy: Make sure all added `print` statements will be `python3` compatible, e.g., `print()` vs `print`
If the rest of the file being modified uses the `"{}".format()` style formatting, rather than `"%x"%val` style, also conform to that

jsturdy (changes requested):

iterativeTrim.py
AndrewLevin: done

jsturdy (dismissed):

mexanick (changes requested):
@AndrewLevin please rebase

AndrewLevin:
> @AndrewLevin please rebase

done

iterativeTrim.py
```diff
@@ -195,6 +197,26 @@ def writeChConfig(chConfig,vfat,vfatID,trimDAC,trimPol,mask,maskReason):
...
+            for event in vfatTree :
+                # Skip masked vfats
+                if (args.vfatmask >> int(event.vfatN)) & 0x1:
+                    continue
+                    
+                # Write CFG_THR_ARM_DAC
+                print('Set link {0} VFAT{1} CFG_THR_ARM_DAC to {2}'.format(args.link,event.vfatN,event.vt1))
+                vfatBoard.setVFATThreshold(chip=int(event.vfatN), vt1=int(event.vt1))
+        except IOError as e:
+            print('{0} does not seem to exist or is not readable'.format(options.filename))
```
jsturdy: One thing I didn't ask before, if this exception is raised, what is the resulting behaviour (it doesn't look like an exit or something like that)?

jsturdy (dismissed):

iterativeTrim.py
AndrewLevin: First, `options.filename` should be `options.vfatConfig`. 

The resulting behavior if the exception is raised is that it will whatever thresholds the VFATs happen to be configured with when it is run.

jsturdy (approved):

PR #284, opened 2019-11-05
[Bugfix] Fix incorrect array size leading to memory corruptions

lpetre-ulb:
## Description

This pull request adapts the results array size in `sbitRateScanAllLinks` to match the changes made in https://github.com/cms-gem-daq-project/cmsgemos/pull/306

### Types of changes

## Motivation and Context

While testing the GLIB, segmentation faults and double free errors occured in `run_scans.py sbitThresh`.

## How Has This Been Tested?

Using seamlessly and `run_scans.py sbitThresh` for both electronics QC (CTP7/12 OHs) and GLIB testing (2 OHs).

## Checklist:


jsturdy:
Does this pose any issues for code that is compiled with different constants, e.g., 

https://github.com/cms-gem-daq-project/ctp7_modules/blob/c80e01207f21afd768758cc51e8d5994496a31df/src/calibration_routines.cpp#L649

lpetre-ulb:
In theory, yes, it does pose issues. In practice, however, we are "protected" by the fact that `xHAL` uses a fixed number of OHs: https://github.com/cms-gem-daq-project/xhal/blob/7d20a6034d4294b97b2ab2331fe4d8d90ae885c4/xhalcore/src/common/rpc_manager/calibration_routines.cc#L274

Moreover, the number of OH is currently set to 12 for both GE1/1 and GE2/1. The problem is that the number of OH per AMC is not a GEM station specific constant, but rather a function of the GEM station, backend board and firmware build. It should be read from the firmware (at module initialization?) or the RPC be compiled with a 3D matrix (GEM station, backend board, firmware build).

Reading the number of OHs at runtime does not suit very well `xHAL` and its stateless C functions. And using different sets of constants at compile time would mean creating different libraries... I think the only true solution is to wait for the templated RPC...

utils/scanUtils.py
```diff
@@ -453,7 +453,7 @@ def sbitRateScanAllLinks(args, rateTree, vfatBoard, chan=128, scanReg="CFG_THR_A
 
     # Make the containers
     nDACValues = (args.scanmax-args.scanmin+1)/args.stepSize
-    arraySize = amcBoard.nOHs * nDACValues
+    arraySize = 12 * nDACValues
     scanDataDAC = (c_uint32 * arraySize)()
     scanDataRate = (c_uint32 * arraySize)()
     scanDataRatePerVFAT = (c_uint32 * (24 * arraySize))() # per VFAT
```
jsturdy, line 459: I suppose that in all upstream places, this `24` is also hard coded, so we don't have any issue there?

jsturdy (approved):

utils/scanUtils.py
lpetre-ulb: That's a good question. It seems that the number of VFATs is *not* hardcoded, but passed as an argument or set a compile time. However, it is safe to use an array which is bigger than necessary.

I don't currently have a setup to test a patch where 24 is replaced with `nVFATs` and wehre `nVFATs` is actually different than 24. However, if the code did work previously with GE2/1, there is no reason it wouldn't work anymore.

PR #285, opened 2019-11-06
Handle bad DAC fit exceptions in testConnectivity.py

AndrewLevin:
This pull request makes the changes to `vfatqc-python-scripts` that go along with the changes to `gem-plotting-tools` in https://github.com/cms-gem-daq-project/gem-plotting-tools/pull/260

## Description
This pull request updates `testConnectivity.py` to handle the new exceptions that are thrown when there is a large chisquare value for a DAC vs ADC fit. The default behavior is to exit, but there is a a flag that can be passed to `testConnectivity.py` that tells it to ignore these exceptions.

### Types of changes

## Motivation and Context
We are mainly using `testConnectivity.py` to run DAC scans now.

## How Has This Been Tested?
Yes, I tested it on the coffin, with and without the option `--acceptBadDACFits`, and found the expected behavior in both cases

## Checklist:

### DIFF
--- a/testConnectivity.py
+++ b/testConnectivity.py
@@ -194,6 +194,8 @@ def testConnectivity(args):
     from gempython.vfatqc.utils.qcutilities import getCardName
     if hasattr(args, 'acceptBadDACBiases') is False: # Accept Bad cases where a VFAT DAC cannot reach correct bias voltage/current
         args.acceptBadDACBiases = False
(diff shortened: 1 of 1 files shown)


jsturdy:
@AndrewLevin can you rebase this?

jsturdy:
> @AndrewLevin can you rebase this?

Actually, it's only one commit, so I feel comfortable doing this in the GUI (though it loses the merge commit)

PR #289, opened 2019-12-18
add iterative trimming with parallel analysis

AndrewLevin:
The script `iterativeTrim.py` is replaced by the script `iterativeTrimParallel.py` which takes an optohybrid mask argument instead of a link number argument.

## Description
In each iteration, the scurves are taken in sequence and then the analysis is performed in parallel using the `multiprocessing` module. Like for `ana_scans.py`, the three options of `light`, `medium`, and `heavy` are used to control the CPU usage of the parallel jobs.

### Types of changes

## Motivation and Context
The scurve analysis actually takes more time than the scurve data taking.

## How Has This Been Tested?
This has been tested on the coffin which has a single link.

## Checklist:


jsturdy:
What is the rationale behind removing/renaming the file, rather than simply updating the behaviour?
I would keep the same file, and just change the internals, especially as there is not a companion non-parallel option after this change...

AndrewLevin:
The rationale was to make it consistent with this file name

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/rpc-playground/sbitThreshScanParallel.py

iterativeTrimParallel.py
```diff
@@ -0,0 +1,593 @@
...
+            continue    
+
+        # Get Channel Register Info
+        if dict_chanRegData is None:
+            setChanRegs = False
+            cArray_trimVal = None
+            cArray_trimPol = None
+        else:
+            setChanRegs = True
+            cArray_trimVal = (c_uint32 * 3072)(*dict_chanRegData[ohN]["ARM_TRIM_AMPLITUDE"])
```
mexanick: the 3072 constant should not be hardcoded, but deduced from them gem type (for GE1/1 it is 3072, for GE2/1 it is 1536, for ME0 - tbd)

iterativeTrimParallel.py
```diff
@@ -0,0 +1,593 @@
...
+    maskTransform = lambda x: (x & 0x1) << 5
+
+    # Make containers for tracking iteration progress
+    scurveStatTypes=[ ("iterN","uint8"),("vfatN","uint8"),("vfatID","uint32"),("avg","f4"),("std","f4"),("max","f4"),("min","f4"),("p2p","f4"),("n_trimmed","uint8") ]
+
+    dict_chanRegArray = {}
+    dict_chanRegArray[0] = {}
+    
+    #initialize bookkeeping information
+    for ohN in range(0,amcBoard.nOHs+1):
```
mexanick: Why do we have `amcBoard.nOHs+1` as a top end of range? This adds an extra iteration (which doesn't mean anything as such OH will always be masked)

mexanick (changes requested):

jsturdy:
> The rationale was to make it consistent with this file name
> 
> https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/rpc-playground/sbitThreshScanParallel.py

Is there any reason we keep around a non-"parallel" `sbitThreshScan`?
If not, then i'd just remove reference to "parallel" from filenames, options, analysis, etc., I've always had a bad smell about this "parallel" jargon that creeps everywhere

iterativeTrimParallel.py
AndrewLevin: updated with `vfatsPerGemVariant`

iterativeTrimParallel.py
AndrewLevin: I don't see any reason for this `+1` either. I fixed this here and in all other places in `vfatqc-python-scripts`.

AndrewLevin:
There is no non-parallel sbitThreshScan in the repository. I removed the "Parallel" from sbitThreshScanParallel and iterativeTrimParallel.

jsturdy (approved):

Issue #295, opened 2020-01-31
Iterative trimmer crashes due to TypeError


AndrewLevin:
The syntax `(c_uint32*vfatsPerGemVariant[gemType]*CHANNELS_PER_VFAT)` here

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/rpc-playground/iterativeTrim.py#L59-L60

does not work.

## Brief summary of issue
The use of the `*` for multiplication is probably not correct here because it is also used to indicate pointer.

### Types of issue

- [x] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior
Iterative trimmer should not crash

## Current Behavior
Iterative trimmer crashes with the following error:
```
Traceback (most recent call last):
  File "/opt/cmsgemos/bin/iterativeTrim.py", line 413, in <module>
    scurveFitResults = iterativeTrim(args,dict_dirPaths,identifier,dict_chanRegData=dict_chanRegArray[iterNum-1], dict_calFiles=None)
  File "/opt/cmsgemos/bin/iterativeTrim.py", line 59, in iterativeTrim
    cArray_trimVal = (c_uint32*vfatsPerGemVariant[gemType]*CHANNELS_PER_VFAT)(*dict_chanRegData[ohN]["ARM_TRIM_AMPLITUDE"])
TypeError: expected c_uint_Array_24 instance, got numpy.uint8
```

## Possible Solution (for bugs)
Define a temporary variable.

PR #296, opened 2020-01-31
Fix c_uint32 array length definition bug

AndrewLevin:
Defines a temporary variable to hold the length of the array instead of performing multiplication inside of the parentheses to calculate the length.

## Description
The syntax used to create an array of `c_uint32`s was not correct because of the additional `*`.

### Types of changes

## Motivation and Context
Resolves https://github.com/cms-gem-daq-project/vfatqc-python-scripts/issues/295

## How Has This Been Tested?
Yes, it works in my virtual environment.

## Checklist:

### DIFF
--- a/iterativeTrim.py
+++ b/iterativeTrim.py
@@ -56,8 +56,8 @@ def iterativeTrim(args,dict_dirPaths,identifier,dict_chanRegData=None,dict_calFi
             setChanRegs = True
             detName = chamber_config[ohKey]
             gemType = detName[:detName.find('-')].lower()
-            cArray_trimVal = (c_uint32*vfatsPerGemVariant[gemType]*CHANNELS_PER_VFAT)(*dict_chanRegData[ohN]["ARM_TRIM_AMPLITUDE"])
-            cArray_trimPol = (c_uint32*vfatsPerGemVariant[gemType]*CHANNELS_PER_VFAT)(*dict_chanRegData[ohN]["ARM_TRIM_POLARITY"])
+            cArray_trimVal = (c_uint32*(vfatsPerGemVariant[gemType]*CHANNELS_PER_VFAT))(*dict_chanRegData[ohN]["ARM_TRIM_AMPLITUDE"])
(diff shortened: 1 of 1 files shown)


mexanick (dismissed):

iterativeTrim.py
```diff
@@ -56,8 +56,9 @@ def iterativeTrim(args,dict_dirPaths,identifier,dict_chanRegData=None,dict_calFi
             setChanRegs = True
             detName = chamber_config[ohKey]
             gemType = detName[:detName.find('-')].lower()
-            cArray_trimVal = (c_uint32*vfatsPerGemVariant[gemType]*CHANNELS_PER_VFAT)(*dict_chanRegData[ohN]["ARM_TRIM_AMPLITUDE"])
-            cArray_trimPol = (c_uint32*vfatsPerGemVariant[gemType]*CHANNELS_PER_VFAT)(*dict_chanRegData[ohN]["ARM_TRIM_POLARITY"])
+            nChannels = vfatsPerGemVariant[gemType]*CHANNELS_PER_VFAT
+            cArray_trimVal = (c_uint32*nChannels)(*dict_chanRegData[ohN]["ARM_TRIM_AMPLITUDE"])
+            cArray_trimPol = (c_uint32*nChannels)(*dict_chanRegData[ohN]["ARM_TRIM_POLARITY"])
             pass
```
jsturdy, line 61: If you want to preserve the semantics of the dimensionality, you can just do, e.g.,
```python
             cArray_trimVal = (c_uint32*(vfatsPerGemVariant[gemType]*CHANNELS_PER_VFAT))(*dict_chanRegData[ohN]["ARM_TRIM_AMPLITUDE"])
```
e.g., put the argument to the `c_uint32` operation inside parentheses, and thus avoid the explicit temporary

AndrewLevin: ok, done

jsturdy: i'd simply squash those two commits, so there's only the one change in the history

AndrewLevin: ok, done

jsturdy (approved):
:+1:

PR #300, opened 2020-02-26
legacy releasing

jsturdy:
Bump `gembuild` submodule to be able to have rpm creation to be the same as other repos, i.e., getting the tag and meta info from the `tag2rel.sh` script

With a tag 2.7.11 one commit behind the build, the change results in:
```
./rpm/gempython_vfatqc-2.7.11-1.0.1.dev.7118576git.centos7.python2.7.noarch.rpm
```
the dirty tagged commit of 2.7.11 results in
```
./rpm/gempython_vfatqc-2.7.11-1.caf96cagit.centos7.python2.7.noarch.rpm
```
compared to the old, hardcoded way
```
./rpm/gempython_vfatqc-2.7.11-centos7.python2.7.x86_64.rpm
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,9 +21,9 @@ PythonModules = ["$(Namespace).$(ShortPackage)", \
 ]
 $(info PythonModules=${PythonModules})
 
-VFATQC_VER_MAJOR=2
-VFATQC_VER_MINOR=4
-VFATQC_VER_PATCH=3
+VFATQC_VER_MAJOR:=$(shell ./config/tag2rel.sh | awk '{split($$0,a," "); print a[1];}' | awk '{split($$0,b,":"); print b[2];}')
+VFATQC_VER_MINOR:=$(shell ./config/tag2rel.sh | awk '{split($$0,a," "); print a[2];}' | awk '{split($$0,b,":"); print b[2];}')
+VFATQC_VER_PATCH:=$(shell ./config/tag2rel.sh | awk '{split($$0,a," "); print a[3];}' | awk '{split($$0,b,":"); print b[2];}')
 
 include $(BUILD_HOME)/$(Project)/config/mfCommonDefs.mk
 include $(BUILD_HOME)/$(Project)/config/mfPythonDefs.mk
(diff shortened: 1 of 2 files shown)


lpetre-ulb (commented):
> Bump `gembuild` submodule to be able to have rpm creation to be the same as other repos, i.e., getting the tag and meta info from the `tag2rel.sh` script

Cursory approval. Small question in addition, is the `gembuild` submodule bumped to the same commit as the other repositories? Is there a legacy branch in `gembuild` which is going to be used as the reference for legacy releases?

lpetre-ulb (approved):

jsturdy:
the different legacy repos are often at different commits, this one lagged behind for some other reasons, i bumped it to the same as `gem-plotting-tools` which still allows things to build without other changes.
it's possible that an even further commit would still build, but at some point the changes become incompatible with how things were being done here...
N.B. the non-legacy branches of all repos should all be on the same commit (i'd probably prefer having them track `master` rather than some fixed commit, i.e., once things are progressing, and the `gembuild` changes are all merged in...)

mexanick (approved):

lpetre-ulb:
Ok, thank you for the precisions.

> the different legacy repos are often at different commits, this one lagged behind for some other reasons, i bumped it to the same as `gem-plotting-tools` which still allows things to build without other changes.
> it's possible that an even further commit would still build, but at some point the changes become incompatible with how things were being done here...

Yeah, let's play it safe for the legacy branches. I would have hope that the breaking commit was easy to identify...

> N.B. the non-legacy branches of all repos should all be on the same commit (i'd probably prefer having them track `master` rather than some fixed commit, i.e., once things are progressing, and the `gembuild` changes are all merged in...)

Agreed, unfortunately there is no easy way to follow a branch in a Git submodule. :-/

jsturdy:
> Agreed, unfortunately there is no easy way to follow a branch in a Git submodule. :-/

Actually, this is no longer true (requires at least `git` 1.8.2+) :+1:, though I don't know how the git(hub|lab) servers work with this...

lpetre-ulb:
Well, AFAIK, it is possible to track a Git branch for each submodule, but the update itself still requires a command like  `git submodule update --remote`. Do you have any other tip?

PR #304, opened 2020-04-15
[feature] Unified docs building for legacy code

jsturdy:
## Description
Uniform building of docs for legacy code:
* on demand creation of `conda` environment for python deps (if specified)
  * `USE_CONDA` can be set at docs building time to flag these behaviours
  * if `USE_CONDA` is not set, the environment from which `make doc` is called needs all dependencies installed
* split source and build directories for `sphinx`
* use `sphinx-autoapi` for python docs automation
  * no need for manually creating the `autodoc` stubs

### Types of changes

## Motivation and Context
Requests for documentation of legacy code

## How Has This Been Tested?
https://test-gemdaq-ci.web.cern.ch/test-gemdaq-ci/docs/api/vfatqc/latest/


mexanick:
is this setup still valid?
https://test-gemdaq-ci.web.cern.ch/test-gemdaq-ci/docs/api/vfatqc/latest/setup.html
also, do we need a section considering the `venv` at P5? I think we are *not going* to support `venv`

jsturdy:
> is this setup still valid?

Don't know, I didn't do any verification of information because I don't know of anything/anyone using this as a source of information (the go-to source is/was the electronics user guide)

> https://test-gemdaq-ci.web.cern.ch/test-gemdaq-ci/docs/api/vfatqc/latest/setup.html
> also, do we need a section considering the `venv` at P5? I think we are _not going_ to support `venv`

It was never supported, but someone wrote it in here anyway :-D

I can gut whatever or everything and leave just the API part

mexanick:
But this documentation **should become** a source of information, so it must be correct. How do you envision checking upon it?

jsturdy:
> But this documentation **should become** a source of information, so it must be correct. How do you envision checking upon it?

That's entirely unclear, given that this legacy code is not going to be supported long

cgalloni:
> That's entirely unclear, given that this legacy code is not going to be supported long

Actually we would like to recommend the usage of the legacy software for remote stands.
And I think we all agreed that the web page should become the official source of documentation and instructions: correct, clear and easier to follow than the current electronics guide.

lpetre-ulb:
I agree with @mexanick and @cgalloni, this documentation should become **the** source of information for legacy code, whether it will be supported for a long time. We still don't know how long it will take for the new software to on on-pair with the features. Multiple test stands (CERN and non-CERN) are probaly going to use legacy software for months to come. If the documentation is completely outdated and misleading, such as in this case, it can simply be removed. Other parts of the documentation must be checked and updated if needed (e.g. information originating from the electronics guide).

jsturdy:
OK, in this case I'll just remove all non-API information and somehow migrate that to the users guide

jsturdy:
* last changes I think have been pushed.
* docs site updated
* docstrings *could* be cleaned up in the python code, because there are various warnings/errors in the building, but they don't really prevent the docs from rendering, they would simply be made nicer

lpetre-ulb (approved):

